PR #1, opened 2018-02-13
 Basic address schema and migrations 

oyeb:
+ Do we really need the following fields?
  * `first_name`
  * `last_name`
+ Should `address_line_2` be a "required field"?
+ Min. length of `address_line_1` is set to 10. Is that ok?

This feature is **blocked by `State`, `Country`**.

### DIFF
--- a/apps/core/lib/core/snitch/address.ex
+++ b/apps/core/lib/core/snitch/address.ex
@@ -11,27 +11,34 @@ defmodule Core.Snitch.Address do
 
   import Ecto.Changeset
 
+  @type t :: %__MODULE__{}
+
   schema "snitch_addresses" do
-    field :first_name, :string
-    field :last_name, :string
-    field :address_line1, :string
-    field :address_line2, :string
-    field :city, :string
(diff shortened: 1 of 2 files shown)


apps/core/lib/core/snitch/address.ex
```diff
@@ -0,0 +1,46 @@
...
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
+
+  schema "snitch_addresses" do
+    # Do we really need first and last name separately in Address??
+    field(:first_name, :string)
+    field(:last_name, :string)
```
pkrawat1: Who does these first_name, last_name referred to

oyeb: In spree, the user's name is simply copied in here. It's an editable field in the UI.

ashish173: Address is always stored in modular form, so they make total sense @oyeb

PR #2, opened 2018-02-13
Added basic User model, with pwd encryption.

oyeb:
* Very limited fields!
* Uses Comeonin's [Argon2](https://hexdocs.pm/argon2_elixir/Argon2.html#hash_pwd_salt/2) password hashing algorithm, interestingly it uses random salts and we don't store salts in DB/anywhere.
* There's unique_index on `:email`.


apps/core/priv/repo/migrations/20180213064751_create_user.exs
```diff
@@ -0,0 +1,46 @@
...
+      add :email, :string
+      add :password_hash, :string, comment: "Generated by argon2_elixir:hash_pwd_salt with default 16 byte salt length."
+      add :is_admin?, :boolean, default: false
+
+      add :sign_in_count, :integer, default: 0
+      add :failed_attempts, :integer, default: 0
+
+      # add :snitch_api_key, :string
+      # add :remember_token, :string
+
```
ashish173: Don't understand why these are even in place commented; either they go in the migration or don't. @oyeb

PR #10, opened 2018-02-21
add custom mix tasks for db reset and seed

### LINKED ISSUE #7: mix task for seeding data
Create a mix task to seed data for the `core` app. 
The `applicaitons` needed to seed data should be loaded only during the seed operation.

arjun289:
added custom mix task for: 
* resetdb 
```elixir
  mix snitch.resetdb
```
* seed db 
```elixir
  mix snitch.seed
```
closes #7 

The commands need to be run from the `root` of the application.


apps/core/lib/mix/tasks/core_db_task.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Mix.Tasks.Core.Resetdb do
```
oyeb: We already have a handy [alias](https://github.com/aviabird/snitch/blob/feature%2Fcustom_mix_task/apps/core/mix.exs#L69) for this.

apps/core/lib/mix/tasks/core_db_task.ex
```diff
@@ -0,0 +1,46 @@
...
+      into: IO.stream(:stdio, :line)
+    )
+  end
+end
+
+defmodule Mix.Tasks.Core.Seed do
+  @moduledoc """
+    Custom mix task to seed db.
+    ```
+      iex> mix core.seed
```
oyeb: This is not run in an IEx shell :grinning:

apps/core/mix.exs
```diff
@@ -25,7 +25,7 @@ defmodule Core.Mixfile do
   def application do
     [
       mod: {Core.Application, []},
-      extra_applications: [:logger, :runtime_tools, :yamerl, :worldly]
+      extra_applications: [:logger, :runtime_tools]
     ]
```
oyeb: Please remove `uuid` and `excountries` from deps and restrict `yamerl` and `worldly` to `only: dev`

apps/core/lib/mix/tasks/core_db_task.ex
oyeb: Also rename this file to maybe `seed.ex`

arjun289: I think we should keep it the same keeping in mind that it will be required in future when our app core is used as a library and we will have to use this task we won't be using alias then.

apps/core/lib/mix/tasks/core_db_task.ex
arjun289: sorry my bad will change this.

oyeb:
Is [this](https://github.com/aviabird/snitch/blob/develop/apps/core/config/config.exs#L4) still required?

apps/core/lib/mix/tasks/core_db_task.ex
oyeb: Hmmm... that's a valid point. Then we better rename this to `snitch.setup` or snitch.db.setup` And seed should be renamed to `snitch.seed`.
I setup a project with snitch as dependency and `core.resetdb` and `core.seed` could not be found.

apps/core/lib/mix/tasks/core_db_task.ex
```diff
@@ -0,0 +1,46 @@
...
+    System.cmd(
+      "mix",
+      ["ecto.migrate"],
+      stderr_to_stdout: true,
+      into: IO.stream(:stdio, :line)
+    )
+  end
+end
+
+defmodule Mix.Tasks.Core.Seed do
```
oyeb: I think each task should be in it's own file too.

arjun289: we can put db related tasks together.

oyeb:
I setup a project with snitch as dependency (from the `feature/custom_mix_task`) and `core.seed` failed with this message:
```sh
** (Mix) No such file: ./apps/core/priv/repo/seeds.exs
```
`core.resetdb` worked after I put this in my config:
```elixir
config :my_snitch_app, ecto_repos: [Core.Repo]
config :core, Core.Repo,
  adapter: Ecto.Adapters.Postgres,
  username: "postgres",
  password: "postgres",
  database: "my_snitch_app_dev",
  hostname: "localhost",
  pool_size: 10
```

oyeb:
This task is very slow and can be sped up by:
1. parallelizing the inserts (chunk data and use Task.async)
2. Using `Ecto.Repo.insert_all/2`

pkrawat1:
Please fix the comments, resolve conflicts

ashish173:
I am closing this PR as it's not going to be merged and is not in line with our next 2-3 month milestones.

PR #12, opened 2018-02-23
Added quantifier domain

pkrawat1:
### Stock Item
- Addition
  - Total count on hand available for a variant. Total count might be negative too. If backorder is set.
  - Active stock items collector. Fetches stock items belonging to active stock locations
  - Added specs for stock item model.
- Modification
  - Migration updated
### Stock Location
 - Addition
    - Basic crud added.
    - Active stock location collector added.
    - Added specs for stock location model.
 - Modification
    - Migration updated for missing fields.


apps/core/lib/core/snitch/domain/stock/quantifier.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Core.Snitch.Domain.Stock.Quantifier do
+  @moduledoc """
+    Interface for handling inventory related business logic
+  """
+  
+  use Core.Snitch.Domain
+
+  def total_on_hand(variant) when is_map(variant), do: total_on_hand(variant.id)
```
oyeb: 1. This will only return the stock quantity total from _"active"_ stock locations. There is no method to obtain the total from _all_ locations. We could make this:
`def total_on_hand(variant_id, only_active \\ true)`
2. Please document these methods.

apps/core/lib/core/snitch/data/model/stock/stock_item.ex
```diff
@@ -29,4 +29,24 @@ defmodule Core.Snitch.Data.Model.Stock.StockItem do
   end
 
   def get_all, do: Schema.Stock.StockItem |> Repo.all()
+
+  def stock_items(variant_id) do
+    variant_id
+    |> stock_items_query
```
oyeb: Missing parens!

PR #14, opened 2018-02-26
Add Payment schema, migrations and models

### LINKED ISSUE #16: `QueryHelper.delete` always returns `{:error, :not_found}`
```elixir
def delete(schema, id, repo) when is_integer(id) do
  with {:ok, instance} <- repo.get(schema, id) do
    delete(schema, instance, repo)
  else
    {:error, :not_found}
  end
end
```
The `with` clause always fails because `Ecto.Repo.get/2)` always returns `Ecto.Schema.t` or `nil` and never a tuple.

oyeb:
* As mentioned in #155285071, we only have a subtype for `CardPayment`
* Added payment related factories.
* `Model.Payment*.update` and `Model.Payment*.delete` functions have been marked as `deprecated` as they are dangerous.
* Parens are not needed on `map`, `non_neg_integer`, etc in `specs`. **As a rule of thumb, DON'T add parens anywhere and let the formatter add them for you.**
* Added `{:error, :not_found}` as a return type of `QueryHelper.delete`, also fixed #16.

### Tasks
* [x] Models should ensure supertype is inserted before subtype
* [x] Tests on Models and Schemas


apps/core/lib/core/snitch/data/schema/payment/payment.ex
```diff
@@ -0,0 +1,63 @@
...
+
+  use Core.Snitch.Data.Schema
+
+  @type t :: %__MODULE__{}
+
+  @payment_types ["chk", "ccd"]
+
+  schema "snitch_payments" do
+    field(:slug, :string)
+    field(:payment_type, :string, size: 3)
```
oyeb: Shouldn't we call this _just_ `:type`?

apps/core/lib/core/snitch/data/schema/payment/payment.ex
```diff
@@ -0,0 +1,63 @@
...
+    field(:amount, Money.Ecto.Composite.Type, default: Money.new(0, :USD))
+    field(:state, :string, default: "pending")
+
+    belongs_to(:payment_method, PaymentMethod)
```
oyeb: Shouldn't we call this _just_ `:method`?

apps/core/lib/core/snitch/data/schema/payment/payment.ex
pkrawat1: yes we can, but type, method might collide with keywords, or may lead to confusion

apps/core/lib/core/snitch/data/model/payment/card_payment.ex
```diff
@@ -0,0 +1,113 @@
...
+    Ecto.Multi.new()
+    |> Ecto.Multi.insert(:payment, payment_changeset)
+    |> Ecto.Multi.run(:card_payment, fn %{payment: payment} ->
+      all_card_params = Map.put(card_params, :payment_id, payment.id)
+      QH.create(Schema.CardPayment, all_card_params, Repo)
+    end)
+    |> Core.Repo.transaction()
+  end
+
+  @deprecated "This is dangerous as it allows changing the amount"
```
oyeb: Why should we allow updating amounts? Should we keep this or remove it?

apps/core/lib/core/snitch/data/model/payment/card_payment.ex
```diff
@@ -0,0 +1,113 @@
...
+
+    Ecto.Multi.new()
+    |> Ecto.Multi.update(:card_payment, card_payment_changeset)
+    |> Ecto.Multi.run(:payment, fn _ ->
+      Model.Payment.update(nil, Map.put(payment_params, :id, card_payment.payment_id))
+    end)
+    |> Core.Repo.transaction()
+  end
+
+  @deprecated "Deletion of payments! Tsk tsk, bad idea sir/madam"
```
oyeb: Why should we allow deleting existing payments? Should we keep this or remove it?

apps/core/lib/core/seed/payment_methods.ex
```diff
@@ -0,0 +1,45 @@
...
+  schema and table.
+
+  ## Roadmap
+
+  Snitch will support "Store Credits", which act like e-wallets for users.
+  """
+  use Core.Snitch.Data.Schema
+
+  def seed!() do
+    methods = [
```
oyeb: Since there is no schema or model for check/cash payments, IMO we should not seed this payment method in this PR.

apps/core/lib/core/snitch/data/model/payment/card_payment.ex
ashish173: As we discussed last time; since this is not going to be used in the flow that we discussed so it's better if we remove it.

apps/core/lib/core/snitch/data/model/payment/card_payment.ex
ashish173: As we discussed last time; since this is not going to be used in the flow that we discussed so it's better if we remove it.

arjun289: If we are providing the option of partial payments, then the user might need to change the amount he/she wants to pay using the card option in that case this might be needed.

apps/core/lib/core/snitch/data/model/payment/card_payment.ex
arjun289: What if the payment type needs to be changed due to some reason or may there is a request from the customer to the admin to modify the payment type. Then update is not covering this modification and a fresh payment will have to be created in that case.

apps/core/lib/core/snitch/data/model/payment/payment_method.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Core.Snitch.Data.Model.PaymentMethod do
+  @moduledoc """
+  PaymentMethod API and utilities.
+
+  Snitch currently supports the follwoing payment methods:
```
arjun289: typo `following`

apps/core/lib/core/snitch/data/schema/order.ex
```diff
@@ -22,7 +22,7 @@ defmodule Core.Snitch.Data.Schema.Order do
     # field :shipping
     # field :payment
```
arjun289: commented fields?

apps/core/test/snitch/data/model/payment/card_payment_test.exs
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  test "delete", context do
+    %{card_payment: card_payment} = context
+    assert {:ok, card_payment} = Model.CardPayment.delete(card_payment.id)
+    assert card_payment.state == "some-state"
+    assert {:error, :not_found} = Model.CardPayment.delete(-1)
+  end
+  
+  defp card_payment(context) do
```
arjun289: This too can be moved to factory.

apps/core/lib/core/snitch/data/model/payment/card_payment.ex
oyeb: Partial payments do not imply that the user should be able to edit the amount. I'm listing all scenarios where an order has more than 1 payment:
1. Order amount is `x` and a user has `k/100 * x` in his store credit. The user _can_ make up to two payments,
    - `k/100 * x` by store-credit
    - `(100 - k)/100 * x` by card/cash/whatever
2. The user has paid the complete order amount (possibly via multiple payments) AND the admin increases the order total. The user will have to make _at least one_ additional payment now for the outstanding amount.`

apps/core/lib/core/snitch/data/schema/order.ex
oyeb: They haven't been added to the migrations. That's why they are commented.

apps/core/lib/core/seed/payment_methods.ex
```diff
@@ -0,0 +1,45 @@
...
+      %{
+        name: "card",
+        code: "ccd",
+        active?: true,
+        inserted_at: Ecto.DateTime.utc(),
+        updated_at: Ecto.DateTime.utc()
+      }
+    ]
+
+    Core.Repo.insert_all(PaymentMethod, methods, on_conflict: :nothing, conflict_target: :code)
```
yogitabang: Kinda new to this, can you explain me `on_conflict` and `conflict_target`.
A short description will do :)

kartikjagdale: Kinda new to this, can you explain me on_conflict and conflict_target.
A short description will do :)

apps/core/lib/core/snitch/address.ex
```diff
@@ -32,7 +32,7 @@ defmodule Core.Snitch.Address do
   @required_fields ~w(first_name last_name address_line_1 city zip_code)a
```
kartikjagdale: I am kinda a skeptical with the variable being named `address_line_1`, IMO  any variable that relies on a numeral to distinguish it from a similar variable needs to be refactored. I think we should change the existing
name to something more specific.
Not exactly related to this but here is the article might give more insights:
Link: http://archive.oreilly.com/pub/post/the_worlds_two_worst_variable.html

apps/core/lib/core/snitch/data/model/line_item.ex
```diff
@@ -4,7 +4,7 @@ defmodule Core.Snitch.Data.Model.LineItem do
   """
   use Core.Snitch.Data.Model
 
-  @spec get(map()) :: Schema.LineItem.t() | nil | no_return
+  @spec get(map) :: Schema.LineItem.t() | nil | no_return
```
kartikjagdale: I didn't get what  the `no_return` does here, can you elaborate a bit.
And I don't get how are letting the `nil` reach till the `get` of Model, shouldn't it be filtered earlier. or not let it reach the db itself.
Think about it, a db query which searches `nil` and then returns nothing.
I am open to any thoughts on this ?
cc: @ashish173  @pkrawat1

apps/core/lib/core/seed/payment_methods.ex
oyeb: The new `Ecto.Repo.insert_all` allows us to do "upserts". Here, we have a unique index on `:code` so inserting a row with the same `:code` results in a DB (conflict) error. By telling ecto how to handle such an error (do `nothing` if conflict was on column `code`), `insert_all` will succeed.

apps/core/lib/core/snitch/data/model/line_item.ex
oyeb: Pls read [Ecto.Repo](https://hexdocs.pm/ecto/Ecto.Repo.html#c:get/3) specs. `no_return` is for dialyzer. There's nothing wrong when we return `nil`. If there is no such record in the DB, you might want to return `:error` or `{:error, :not_found}`:
1. Both `:error` and `nil` are atoms, so it makes no difference. `nil` is clearly better.
2. Matching on `nil` is much easier than `{:error, :not_found}`, moreover the only "error" a get might encounter is "not_found", and there's no need to explicitly pass that atom around.

apps/core/lib/core/snitch/data/model/line_item.ex
```diff
@@ -43,7 +43,7 @@ defmodule Core.Snitch.Data.Model.LineItem do
   #Money<:USD, 19.98000000>
   ```
   """
-  @spec update_price_and_totals([map()]) :: [map()]
+  @spec update_price_and_totals([map]) :: [map]
```
kartikjagdale: The whole Logic looks good, ThumbsUp 👍  for using `Stream.map`.
However I have yet another approach, correct me if I am wrong.
Here we are calculating the `unit_selling_prices` in `update_price_and_totals`
it will return a `Enumerable`.
Which again we are passing to `set_price_and_total`, which takes exactly one `line_item` where on line no. `62`
we are using it(`unit_selling_prices`) to retrieve `unit_price` for one `variant_id`.

Could we have had a separate method which would calculate the `unit_price` for one `line_item` there and there itself. and we just filter out all the `nil` line_times
in the `update_price_and_totals`.
Something just like below:
For `update_price_and_totals`:
line_items |> Enum.map(set_price_and_total())

and calculate the `unit_price` inside another separate method which would be called from `set_price_and_total`

You see the responsibility of `update_price_and_totals` should be only to update the price and details and not calculate `unit_selling_prices`.

Any thoughts or alternate methods on this are appreciated. 😄 
cc: @ashish173  @pkrawat1

oyeb: @kartikjagdale Could you pls open an issue on line_items and price computation because this PR is about payments.

apps/core/lib/core/snitch/data/model/order.ex
```diff
@@ -20,14 +20,14 @@ defmodule Core.Snitch.Data.Model.Order do
...
+          {:ok, Schema.Order.t()} | {:error, Ecto.Changeset.t()} | {:error, :not_found}
   def delete(id_or_instance) do
     QH.delete(Schema.Order, id_or_instance, Repo)
   end
 
-  @spec get(map) :: Schema.Order.t() | nil | no_return
-  def get(query_fields) do
-    QH.get(Schema.Order, query_fields, Repo)
+  @spec get(map | non_neg_integer) :: Schema.Order.t() | nil | no_return
+  def get(query_fields_or_primary_key) do
```
kartikjagdale: why was there a need for change in the name, I guess `query_fields` was self explanatory it can also contain a primary key.
IMO here the name kinda of states that there will be a `query_fields`(i.e a map) or a `primary_key`(i.e a integer(a singular entity)).
Please let me know your thoughts on this. 😄 
cc: @ashish173  @pkrawat1

apps/core/lib/core/snitch/data/model/payment/card_payment.ex
```diff
@@ -0,0 +1,95 @@
...
+  @spec create(String.t(), non_neg_integer(), map, map) ::
+          {:ok, %{card_payment: Schema.CardPayment.t(), payment: Schema.Payment.t()}}
+          | {:error, Ecto.Changeset.t()}
+  def create(slug, order_id, payment_params, card_params) do
+    payment = struct(Schema.Payment, payment_params)
+    card_method = Model.PaymentMethod.get_card()
+
+    more_payment_params = %{
+      order_id: order_id,
+      payment_type: "ccd",
```
kartikjagdale: Can we a constant file for this from which we can pick using the key ?
its kinda getting repeated everywhere, would be difficult to make change everywhere in future if there is a need.  😄

apps/core/lib/core/snitch/data/model/order.ex
oyeb: > Here the name kinda of states that there will be a query_fields(i.e a map) or a primary_key(i.e a integer(a singular entity)).

That is exactly what it is. Either a `map`, or an `integer`. See the spec on line 28:
`@spec get(map | non_neg_integer) :: Schema.Order.t() | nil | no_return`

apps/core/lib/core/snitch/data/model/payment/card_payment.ex
```diff
@@ -0,0 +1,95 @@
...
+      payment_method_id: card_method.id,
+      slug: slug
+    }
+
+    payment_changeset = Schema.Payment.changeset(payment, more_payment_params, :create)
+
+    Ecto.Multi.new()
+    |> Ecto.Multi.insert(:payment, payment_changeset)
+    |> Ecto.Multi.run(:card_payment, fn %{payment: payment} ->
+      all_card_params = Map.put(card_params, :payment_id, payment.id)
```
kartikjagdale: I am not sure if it is already done before passing to this method,
but could we have had a `changeset` for  `card_params`/ `all_card_params`  before running `create` query  on `CardPayment`.
Any Thoughts on this ?
cc: @ashish173  @pkrawat1

apps/core/lib/core/snitch/data/model/payment/payment_method.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  There's no separate schema for such payments as they are completely expressed
+  by the fields in `Core.Snitch.Data.Model.Payment`.
+  """
+  use Core.Snitch.Data.Model
+
+  @spec create(String.t(), String.t(), boolean()) ::
+          {:ok, Schema.PaymentMethod.t()} | {:error, Ecto.Changeset.t()}
+  def create(name, code, is_active? \\ true) do
+    pm = %{name: name, code: code, active?: is_active?}
```
kartikjagdale: `pm` ☹️ 
`payment_method_params` 😃

apps/core/lib/core/snitch/data/model/payment/payment_method.ex
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  @spec get(map | non_neg_integer) :: Schema.PaymentMethod.t() | nil | no_return
+  def get(query_fields_or_primary_key) do
+    QH.get(Schema.PaymentMethod, query_fields_or_primary_key, Repo)
+  end
+
+  @spec get_card() :: Schema.PaymentMethod.t() | nil | no_return
+  def get_card() do
+    get(%{code: "ccd"})
```
kartikjagdale: Again just a reminder for file containing string constant. 😄

apps/core/lib/core/snitch/data/schema/payment/payment.ex
```diff
@@ -0,0 +1,84 @@
...
+  defp validate_discriminator(changeset, _, _), do: changeset
+
+  defp validate_amount(%{valid?: true} = changeset, key) do
+    {_, amount} = fetch_field(changeset, key)
+
+    if Decimal.cmp(amount.amount, Decimal.new(0)) != :lt do
+      changeset
+    else
+      changeset
+      |> add_error(:amount, "must be greater than 0", validation: :amount)
```
kartikjagdale: Again a constant file with error messages.

apps/core/test/snitch/data/model/payment/card_payment_test.exs
```diff
@@ -0,0 +1,43 @@
+defmodule Core.Snitch.Data.Model.CardPaymentMethodTest do
+  use ExUnit.Case, async: true
+  use Core.Snitch.Data.Model
+  import Core.Snitch.Factory
+
+  setup :checkout_repo
+  setup :a_user_and_address
```
kartikjagdale: `a_user_and_address` ?
naming...

apps/core/lib/core/snitch/data/model/payment/card_payment.ex
oyeb: `QH.create(Schema.CardPayment, all_card_params, Repo)` (line 45) builds the changeset.

apps/core/lib/core/snitch/data/model/payment/card_payment.ex
oyeb: Sure, that makes a lot of sense.
1. `seed/payment_method.exs` Will read that file and insert accordingly.
2. Code in other places (like here) can do the same.

IMO this is file would be like a config file, so let's add it later (I opened #20 on this)

apps/core/lib/core/snitch/data/model/line_item.ex
oyeb: I've documented why `update_price_and_totals` is like how it is:
>   ## Note
>  Selling prices of all `LineItem`s are fetched from the DB in a single query.

oyeb:
@kartikjagdale  Thanks for the review, and I request you to open issues related to `line_item`, `address` or `order`. I have addressed all comments related to `payments`.

PR #19, opened 2018-03-05
Switch to CodeCov and speed up build

oyeb:
* [inch_ex](https://github.com/rrrene/inch_ex/blob/master/lib/inch_ex/reporter/remote.ex#L45) skips itself on PR builds, which means there's no need to build the project in `docs` env. In fact, Inch reports will now be made only if the commit is for `master` and `develop` branches.
* By sandwiching `mix format --check-formatted` by `set -e` and `set +e`, the build fails early. Otherwise, the tests ran even though the build was failing.
* CodeCov is much better than coveralls.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,22 @@
+coverage:
+  precision: 2
+  round: down
+  range: "60...90"
+
+  status:
+    project:
+      default: on
+    patch:
+      default: on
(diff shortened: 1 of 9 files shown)


oyeb:
Our new test coverage host: https://codecov.io/gh/aviabird/snitch/

PR #28, opened 2018-03-08
Credo issues fixed

jyotigautam:
Credo issues resolved most of which were:
1. Aliasing nested modules.
2. Use normal function call syntax instead of a single pipe operator.
3. Correcting order of Module directives and special attributes.
4. Whitespace issue

### DIFF
--- a/apps/snitch_core/config/test.exs
+++ b/apps/snitch_core/config/test.exs
@@ -8,5 +8,3 @@ config :snitch_core, Snitch.Repo,
   database: "snitch_test",
   hostname: "localhost",
   pool: Ecto.Adapters.SQL.Sandbox
-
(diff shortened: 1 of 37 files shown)


apps/snitch_core/config/test.exs
```diff
@@ -8,5 +8,3 @@ config :snitch_core, Snitch.Repo,
   database: "snitch_test",
   hostname: "localhost",
   pool: Ecto.Adapters.SQL.Sandbox
-
-config :logger, level: :info
```
oyeb: Thanks for removing this! I had unintentionally committed this :sweat_smile:

ashish173: 👍🏼

PR #32, opened 2018-03-11
Pkrawat/stock/movement

### LINKED ISSUE #31: Add Stock Movement for stocks
- Adds schema, model and test cases for stock movements.
- Stock movement is never updated or deleted. So I have not added update/delete methods model.

pkrawat1:
Stock Movements
Fixes #31


apps/snitch_core/lib/core/data/schema/stock/stock_movement.ex
```diff
@@ -0,0 +1,45 @@
...
+
+    belongs_to(:stock_item, StockItem)
+
+    timestamps()
+  end
+
+  @quantity_limits %{min: round(:math.pow(-2, 31)), max: round(:math.pow(2, 31) - 1)}
+  @create_fields ~w(quantity stock_item_id)a
+
+  def create_fields, do: @create_fields
```
oyeb: No use of this function.

apps/snitch_core/lib/core/data/schema/stock/stock_movement.ex
```diff
@@ -0,0 +1,45 @@
...
+
+  @doc """
+  Stock Movements and Stock Transfers are only created.
+  Inorder to update them we can create another stock movement to reverse its effects.
+  """
+  @spec changeset(__MODULE__.t(), map, atom) :: Ecto.Changeset.t()
+  def changeset(instance, params, operation \\ :create)
+  def changeset(instance, params, :create), do: do_changeset(instance, params, @create_fields)
+
+  defp do_changeset(instance, params, fields, optional \\ []) do
```
oyeb: No need for this function either since there is only one changeset function.

apps/snitch_core/lib/core/data/schema/stock/stock_movement.ex
```diff
@@ -0,0 +1,45 @@
...
+  """
+  use Snitch.Data.Schema
+  use Snitch.Data.Schema.Stock
+
+  @type t :: %__MODULE__{}
+
+  schema "snitch_stock_movements" do
+    field(:quantity, :integer, default: 0)
+    field(:action, :string)
+    field(:originator_type, :string)
```
oyeb: What is an `originator`? Do you mean `origin`? Please add a `@typedoc`.
These fields can never be set because your `cast` will throw them out.

apps/snitch_core/lib/core/data/schema/stock/stock_movement.ex
```diff
@@ -0,0 +1,45 @@
...
+  Model to track inventory movement in location
+  """
+  use Snitch.Data.Schema
+  use Snitch.Data.Schema.Stock
+
+  @type t :: %__MODULE__{}
+
+  schema "snitch_stock_movements" do
+    field(:quantity, :integer, default: 0)
+    field(:action, :string)
```
oyeb: What is an `action`? This can never be set because the `cast` will throw it out.

apps/snitch_core/lib/core/data/schema/stock/stock_movement.ex
```diff
@@ -0,0 +1,45 @@
...
+
+  def create_fields, do: @create_fields
+
+  @doc """
+  Stock Movements and Stock Transfers are only created.
+  Inorder to update them we can create another stock movement to reverse its effects.
+  """
+  @spec changeset(__MODULE__.t(), map, atom) :: Ecto.Changeset.t()
+  def changeset(instance, params, operation \\ :create)
+  def changeset(instance, params, :create), do: do_changeset(instance, params, @create_fields)
```
oyeb: The `:create` is required because of the QueryHelper, I believe? Then there's no need to make the `operation` optional -- who else will call this changeset function?

apps/snitch_core/test/data/model/stock/stock_movement_test.exs
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  describe "create/4" do
+    test "Fails for INVALID stock item id" do
+      assert {:error, changeset} = StockMovementModel.create(1, -1)
+      refute changeset.valid?
+      assert %{stock_item_id: ["does not exist"]} = errors_on(changeset)
+    end
+
+    test "Fails with invalid quantity", context do
```
oyeb: No need to test the capability of `cast`. Don't slow down our test suite, it's getting big very fast!

apps/snitch_core/test/data/model/stock/stock_movement_test.exs
```diff
@@ -0,0 +1,75 @@
...
+    end
+
+    test "Inserts with valid attributes", context do
+      %{stock_item: stock_item} = context
+      assert {:ok, stock_movement} = StockMovementModel.create(1, stock_item.id)
+      assert stock_movement.stock_item_id == stock_item.id
+    end
+  end
+
+  describe "get/1" do
```
oyeb: I think we have agreed to **not test** `get`, `get_all`, `delete` if all they do is delegate the work to `QueryHelper`. Please remove these tests. We can probably devise a way to test the QueryHelper methods separately.

apps/snitch_core/test/support/factory/stock.ex
```diff
@@ -31,6 +31,16 @@ defmodule Snitch.StockFactory do
           stock_location: build(:stock_location)
         }
       end
+
+      def stock_movement_factory do
+        %StockMovement{
+          quantity: 1,
+          action: "",
```
oyeb: Please fill these fields with some example text.

apps/snitch_core/lib/core/data/model/stock/stock_movement.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Snitch.Data.Model.StockMovement do
+  @moduledoc """
+  """
+
+  use Snitch.Data.Model
+  alias Snitch.Data.Schema.StockItem, as: StockItemSchema
+  alias Snitch.Data.Schema.StockMovement, as: StockMovementSchema
+
+  @spec create(integer, String.t()) :: {:ok, Ecto.Schema.t()} | {:error, Ecto.Changeset.t()}
+  def create(quantity, stock_item_id) do
```
oyeb: There's no way to set `originator_id`, `originator_type` and `action` field via this method. This module has no docs.

pkrawat1: Keep it as is, until I start logical parts of it. Since It will be needing originator type, action, originator id logic. Currently implemented basic crud for fields required by spree. They never set these fields when creating. It will be a separate feature, related to stock handling(coordinator, packages etc).

apps/snitch_core/test/support/factory/stock.ex
pkrawat1: not checked the logic behind it. Will add it later

apps/snitch_core/test/data/model/stock/stock_movement_test.exs
pkrawat1: ok

apps/snitch_core/test/data/model/stock/stock_movement_test.exs
pkrawat1: Let's say someone changes the error message for quantity, without testing it. Then this could catch it.

apps/snitch_core/lib/core/data/schema/stock/stock_movement.ex
pkrawat1: will be added when originator code is added. Originator can be stock location. or any other schema. But not sure. Haven't started with logics yet

apps/snitch_core/test/data/model/stock/stock_movement_test.exs
oyeb: How would someone change the error message created by `cast`, `cast` takes no such argument?

PR #35, opened 2018-03-12
Set up zones for ShipmentMethod

oyeb:
[**Pivotal Tracker story**](https://www.pivotaltracker.com/story/show/155785133)

### Schema
Zones are of two kinds (`State`, and `Country`), and this is another **non-exclusive** subtype-supertype relationship.

I've chosen to not keep schemas for `StateZone` and `CountryZone` subtype, as there's no table backing them. I considered making them `embedded_schema`s, but we can't put associations in them (can't do `state_zone_struct.members`).

### Models

Conversely, there's no model for `Zone`, instead we have models for `StateZone` and `CountryZone`.
`update/3` has a gotcha: there's no way to add/remove single members from the list.
  - This is because the UI element used to update the member list will definitely have the "new" member list.
  - `update` reduces the number of inserts and deletes to a minimum.

### Tasks

* [x] Speed up tests by using fixtures for state and country, not seed.

### DIFF
--- a/apps/snitch_core/lib/core/data/model/zone/country_zone.ex
+++ b/apps/snitch_core/lib/core/data/model/zone/country_zone.ex
@@ -1,0 +1,131 @@
+defmodule Snitch.Data.Model.CountryZone do
+  @moduledoc """
+  CountryZone API
+  """
+  use Snitch.Data.Model
+
+  import Ecto.Query
+
+  alias Ecto.Multi
+  alias Snitch.Data.Schema.{CountryZoneMember, Zone, Country}
+
+  @doc """
+  Creates a new country `Zone` whose members are `country_ids`.
+
+  `country_ids` is a list of primary keys of the `Snitch.Data.Schema.Country`s that
+  make up this zone. Duplicate IDs are ignored.
+  """
+  @spec create(String.t(), String.t(), [non_neg_integer]) :: term
+  def create(name, description, country_ids) do
+    zone_params = %{name: name, description: description, zone_type: "C"}
+    zone_changeset = Zone.changeset(%Zone{}, zone_params, :create)
+
+    Multi.new()
+    |> Multi.insert(:zone, zone_changeset)
+    |> Multi.run(:members, fn %{zone: zone} -> insert_members(country_ids, zone) end)
+    |> Repo.transaction()
+    |> case do
+      {:ok, %{zone: zone}} -> {:ok, zone}
+      error -> error
+    end
+  end
+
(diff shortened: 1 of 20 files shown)


apps/snitch_core/lib/core/data/model/zone/country_zone.ex
```diff
@@ -0,0 +1,131 @@
...
+    QH.delete(Zone, id_or_instance, Repo)
+  end
+
+  @spec get(map | non_neg_integer) :: Zone.t() | nil
+  def get(query_fields_or_primary_key) do
+    QH.get(Zone, query_fields_or_primary_key, Repo)
+  end
+
+  @spec get_all() :: [Zone.t()]
+  def get_all, do: Repo.all(from(z in Zone, where: z.zone_type == "S"))
```
pkrawat1: what is this `S` type zone and why default scope to it ?

oyeb: Good catch! This should have been `z.zone_type == "C"`.
> why default scope to it?
I'm sorry I don't understand what you mean.

pkrawat1: In rails we some time add default scope, which is like a default where query to get all the records in DB.

Here in this cases `z.zone_type == "C"` is a default scope to which changes the definition of get_all as in get all C type country.

pkrawat1: how about a diff name to function. Like `CountryZone.zone_type_c`

pkrawat1: Now got it. LGTM 👍

apps/snitch_core/lib/core/data/model/zone/country_zone.ex
```diff
@@ -0,0 +1,131 @@
...
+  alias Ecto.Multi
+  alias Snitch.Data.Schema.{CountryZoneMember, Zone, Country}
+
+  @doc """
+  Creates a new state `Zone` whose members are `country_ids`.
+
+  `country_ids` is a list of primary keys of the `Snitch.Data.Schema.Country`s that
+  make up this zone. Duplicate IDs are ignored.
+  """
+  @spec create(String.t(), String.t(), [non_neg_integer]) :: term
```
oyeb: The return type is very lengthy to type so I skipped it. The `Ecto.Multi` specs also skip it :grin:

PR #44, opened 2018-03-28
Fix Order model update

### LINKED ISSUE #43: Order should allow zero LineItems?
~~Creating~~ Updating an order ~~without~~ by removing its `LineItem`s fails because of the way we compute `order.total`:
```
** (Enum.EmptyError) empty error
    (elixir) lib/enum.ex:1855: Enum.reduce/2
    (snitch_core) lib/core/data/schema/order.ex:92: Snitch.Data.Schema.Order.compute_totals/1
    (snitch_core) lib/core/tools/helpers/query_helper/query_helper.ex:37: Snitch.Tools.QueryHelper.update/4
```
This is also related to the way we handle money.

1. The order total is the sum of `line_item.total` (if they are all in the same currency, the other case is not handled).
2. Order schema sets the default price to `#Money<0, :USD>`, but why `USD`?
    a. We should probably pick the "default" curreny from application config.
    b. Does it matter which currency it is? Zero in all currencies is the same :smile: 
3. When we have no `line_items`, The `order.total` must be set to zero... but in which currency?

There are 2 issues here:
1. Should we allow setting an Order's line_items to `[]`?
2. Decide what to do about the default currency.

oyeb:
* `LineItem` costs were (re)computed everytime we called `Model.Order.update`, with this PR the (re)computation is done only when `line_items` are actually changed.
* `Order` could not be created without the addresses, with this PR they can be created without billing and shipping address (to keep in line with the checkout flow).
* Add docs and tests for `Model.Order`
* line_items can be emptied, set up default currency
  - Fixes #43
  - But an "empty" order cannont be created.

## New `Tools`
* Add `money` helper module to generate `zero` monies.
* With this PR, the `Order` schema no longer defines "defaults" for money fields because they must be explicitly calculated. The DB will raise a not-null error if they are not set, so we are super safe.
* Tests that use Application config can never run async, so we made a mock (using [`Mox`](https://github.com/plataformatec/mox))

* `Snitch.Tools.Defaults` is a util to read various defaults from the config, and since it is a `behaviour`, it can fetch them from the `Application` config, system environ, remote server, anything.
* Because it defines a `behaviour` (and also implements it)
    * `Defaults` is tested by actually calling `Application`.
    * `Snitch.Tools.Money` uses `Defaults` in `dev` and `prod` envs but in `test` it uses `Snitch.Tools.DefaultsMock`.
    * `DefaultsMock` is defined in `test/support/mocks.ex`
    * From tests, it is possible to call `Defaults` as well as `DefaultsMock`.


apps/snitch_core/test/data/model/line_item_test.exs
```diff
@@ -29,6 +38,23 @@ defmodule Snitch.Data.Model.LineItemTest do
     end
   end
 
+  describe "compute_total/1 with empty list" do
+    setup do: Application.put_env(:snitch_core, :core_config_app, :snitch)
```
pkrawat1: Add an `on_exit`

apps/snitch_core/lib/core/tools/helpers/money.ex
```diff
@@ -0,0 +1,62 @@
...
+
+  ## Note
+  Makes use of `Money.new/2` and this function won't `raise` unless application is
+  not configured properly.
+  """
+  @spec zero(atom) :: Money.t() | {:error, term}
+  def zero(currency \\ nil)
+
+  def zero(nil) do
+    {:ok, config_app} = Application.fetch_env(:snitch_core, :core_config_app)
```
pkrawat1: Use a `with` here as well.

ashish173:
@oyeb can you resolve conflicts and get it merged if it's ready?

apps/snitch_core/lib/core/data/model/order.ex
```diff
@@ -6,19 +6,118 @@ defmodule Snitch.Data.Model.Order do
...
+  a `line_items`, then **all previous line-items will be deleted!**
+
+  ### Retain previous `LineItem`s
+
+  If you wish to retain the line-items, you must pass a list of maps with the
+  line-item `:id`s, like so:
+
+  ```
+  order # this is the order you wish to update, and `:line_items` are preloaded
+  line_items = Enum.reduce(order.line_items, [], fn x, acc ->
```
oyeb: Wow, that's complicated. Need help in finding an easier way.
It doesn't matter whether you wish to update the `line_items` or not, the `params` must have a `:line_items` key AND the value MUST be a list of maps _not structs_.

The suggested `reduce` is **not trivial**, another option is to call `update/2` like this:
```elixir
same_old_line_items_as_maps = Enum.map(order.line_items, &Map.from_struct/1)

Model.Order.update(%{slug: "foobar", line_items: same_old_line_items_as_maps}, order)
```

apps/snitch_core/lib/core/data/model/line_item.ex
```diff
@@ -55,6 +56,21 @@ defmodule Snitch.Data.Model.LineItem do
     Enum.map(line_items, &set_price_and_total(&1, unit_selling_prices))
   end
 
+  @doc """
+  Returns the item total for given `line_items`.
+
+  If the list is empty, the call is delegated to `MoneyTools.zero/1`.
+  """
+  @spec compute_total([LineItem.t()]) :: Money.t()
+  def compute_total([]), do: MoneyTools.zero!()
```
arjun289: Can this be done using `defdelegate`? just a suggestion

oyeb: No,
1. We're trying to match the empty list, you can't specify args in a `defdelegate` macro.
2. Moreover, the `spec` of both functions doesn't match, a requirement for `defdelegate`.

apps/snitch_core/lib/core/data/model/order.ex
```diff
@@ -6,19 +6,118 @@ defmodule Snitch.Data.Model.Order do
...
+
+  ## Caution!
+
+  The `line_items` are "casted" with the order and if `params` does not include
+  a `line_items`, then **all previous line-items will be deleted!**
+
+  ### Retain previous `LineItem`s
+
+  If you wish to retain the line-items, you must pass a list of maps with the
+  line-item `:id`s, like so:
```
arjun289: Can we handle this on our end? Since we have the  `order id`, we pull it out from db  then update the list with params we received and not asking the user to explicitly enter `ids` of `lineitems` to be preserved. Again just a suggestion.

apps/snitch_core/lib/core/tools/defaults.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Snitch.Tools.Defaults do
```
arjun289: :heart:

apps/snitch_core/lib/core/data/model/line_item.ex
arjun289: ohkk

PR #45, opened 2018-03-30
Added Card Schema

gopalshimpi:
1. Added schema for credit card.
2. Generated Migration for card (snitch_cards).
3. Generated Migration for card_payments (belongs to card).
4. Added factory for card.

### DIFF
--- a/apps/snitch_core/lib/core/data/schema/card.ex
+++ b/apps/snitch_core/lib/core/data/schema/card.ex
@@ -12,11 +12,10 @@ defmodule Snitch.Data.Schema.Card do
     field(:brand, :string)
     field(:is_disabled, :boolean, default: false)
     field(:number, :string)
+    field(:card_name, :string)
 
-    field(:card_name, :string, virtual: true)
-    field(:first_name, :string, virtual: true)
-    field(:last_name, :string, virtual: true)
     field(:last_digits, :string, virtual: true)
+    field(:first_six_digits, :string, virtual: true)
 
     belongs_to(:user, User)
 
@@ -25,60 +24,41 @@ defmodule Snitch.Data.Schema.Card do
   end
 
   @required_fields ~w(name_on_card brand user_id month year number)a
-  @update_fields ~w(is_disabled)a
-  @spec changeset(__MODULE__.t(), map, :create | :update) :: Ecto.Changeset.t()
-  def changeset(card, params, action)
+  @optional_fields ~w(card_name)a
+  @update_fields ~w(is_disabled card_name)a
 
-  def changeset(card, params, :create) do
+  @cast_fields @required_fields ++ @optional_fields ++ @update_fields
+  @spec changeset(__MODULE__.t(), map, :create | :update) :: Ecto.Changeset.t()
+  def changeset(card_changeset, params, action)
+
+  def changeset(card_changeset, params, :create) do
     %{year: current_year, month: current_month} = DateTime.utc_now()
 
-    # for now only for VISA cards.
-
-    card
-    |> cast(params, @required_fields)
+    card_changeset
+    |> cast(params, @cast_fields)
     |> validate_required(@required_fields)
-    |> build_card
     |> validate_number(
       :month,
       less_than_or_equal_to: 12,
       greater_than_or_equal_to: current_month
     )
     |> validate_number(:year, greater_than_or_equal_to: current_year)
-    |> validate_format(:number, ~r/^4[0-9]{12}(?:[0-9]{3})?$/)
+    |> validate_format(:number, ~r/^\d{16}$/)
     |> foreign_key_constraint(:user_id)
+    |> mask_card
   end
 
-  def changeset(card, params, :update) do
-    cast(card, params, @update_fields)
+  def changeset(card_changeset, params, :update) do
+    cast(card_changeset, params, @update_fields)
   end
 
-  def build_card(card) do
-    card
-    |> split_name(Map.has_key?(card.changes, :name_on_card))
-    |> extract_last_digits(
-      Map.has_key?(card.changes, :card_number),
-      Map.has_key?(card.changes, :brand)
-    )
+  def mask_card(card_changeset) do
+    if Map.has_key?(card_changeset.changes, :card_number) do
+      card_changeset
+      |> put_change(:last_digits, String.slice(card_changeset.changes.card_number, -4..-1))
+      |> put_change(:first_six_digits, String.slice(card_changeset.changes.card_number, 0..5))
+    else
+      card_changeset
+    end
   end
-
-  defp split_name(card, true) do
-    [first_name | last_name] = String.split(card.changes.name_on_card, " ")
-    last_name = List.last(last_name)
-
-    card
-    |> put_change(:first_name, first_name)
-    |> put_change(:last_name, last_name)
-  end
-
-  defp split_name(card, false), do: card
-
-  defp extract_last_digits(card, true, true) do
-    card
-    |> put_change(:last_digits, String.slice(card.changes.card_number, -4..-1))
-    |> put_change(:card_name, "#{card.changes.brand} ending in #{card.changes.last_digits}")
-  end
-
-  defp extract_last_digits(card, false, false), do: card
-  defp extract_last_digits(card, true, false), do: card
-  defp extract_last_digits(card, false, true), do: card
 end
--- a/apps/snitch_core/lib/core/data/schema/card.ex
+++ b/apps/snitch_core/lib/core/data/schema/card.ex
@@ -25,9 +25,6 @@ defmodule Snitch.Data.Schema.Card do
   @required_fields ~w(last_digits name_on_card brand user_id month year)a
   @optional_fields ~w(card_name address_id)a
 
-  @current_year DateTime.utc_now() |> Map.fetch!(:year)
-  @current_month DateTime.utc_now() |> Map.fetch!(:month)
-
   @spec changeset(__MODULE__.t(), map, :create | :update) :: Ecto.Changeset.t()
   def changeset(card, params, action)
 
@@ -38,10 +35,10 @@ defmodule Snitch.Data.Schema.Card do
     |> validate_number(
       :month,
       less_than_or_equal_to: 12,
-      greater_than_or_equal_to: @current_month
+      greater_than_or_equal_to: DateTime.utc_now() |> Map.fetch!(:month)
     )
-    |> validate_number(:year, greater_than_or_equal_to: @current_year)
-    |> validate_length(:last_digits, is: 4)
+    |> validate_number(:year, greater_than_or_equal_to: DateTime.utc_now() |> Map.fetch!(:year))
+    |> validate_format(:last_digits, ~r/^\d{4}$/)
     |> foreign_key_constraint(:address_id)
     |> foreign_key_constraint(:user_id)
   end
--- a/apps/snitch_core/lib/core/data/schema/card.ex
+++ b/apps/snitch_core/lib/core/data/schema/card.ex
@@ -1,4 +1,7 @@
 defmodule Snitch.Data.Schema.Card do
+  @moduledoc """
+  """
+
   use Snitch.Data.Schema
 
   alias Snitch.Data.Schema.{CardPayment, User}
@@ -15,7 +18,7 @@ defmodule Snitch.Data.Schema.Card do
     field(:card_name, :string)
 
     field(:last_digits, :string, virtual: true)
-    field(:first_six_digits, :string, virtual: true)
+    field(:first_digits, :string, virtual: true)
 
     belongs_to(:user, User)
 
@@ -24,17 +27,23 @@ defmodule Snitch.Data.Schema.Card do
   end
 
   @required_fields ~w(name_on_card brand user_id month year number)a
-  @optional_fields ~w(card_name)a
   @update_fields ~w(is_disabled card_name)a
+  @cast_fields @required_fields ++ @update_fields
 
-  @cast_fields @required_fields ++ @optional_fields ++ @update_fields
+  @doc """
+  Returns a `Card` changeset.
+
+  The `card_number` should be a string of digits whose length is between 8 to 19
+  digits (inclusive), according to [ISO/IEC
+  7812](https://www.iso.org/obp/ui/#iso:std:iso-iec:7812:-1:ed-5:v1:en)
+  """
   @spec changeset(__MODULE__.t(), map, :create | :update) :: Ecto.Changeset.t()
-  def changeset(card_changeset, params, action)
+  def changeset(card, params, action)
 
-  def changeset(card_changeset, params, :create) do
+  def changeset(%__MODULE__{} = card, params, :create) do
     %{year: current_year, month: current_month} = DateTime.utc_now()
 
-    card_changeset
+    card
     |> cast(params, @cast_fields)
     |> validate_required(@required_fields)
     |> validate_number(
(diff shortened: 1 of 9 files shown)


apps/snitch_core/lib/core/data/schema/card.ex
```diff
@@ -0,0 +1,48 @@
...
+    belongs_to(:user, User)
+
+    has_many(:card_payments, CardPayment)
+    timestamps()
+  end
+
+  @required_fields ~w(last_digits name_on_card brand user_id month year)a
+  @optional_fields ~w(card_name address_id)a
+
+  @current_year DateTime.utc_now() |> Map.fetch!(:year)
```
oyeb: These cannot be module attributes as their values will be compile-time constants. Please compute them at runtime:
```elixir
def changeset(card, params, :create) do
  %{year: current_year, month: current_month} = DateTime.utc_now()
  ...
end
```

apps/snitch_core/lib/core/data/schema/card.ex
```diff
@@ -0,0 +1,48 @@
...
+    card
+    |> cast(params, @required_fields ++ @optional_fields)
+    |> validate_required(@required_fields)
+    |> validate_number(
+      :month,
+      less_than_or_equal_to: 12,
+      greater_than_or_equal_to: @current_month
+    )
+    |> validate_number(:year, greater_than_or_equal_to: @current_year)
+    |> validate_length(:last_digits, is: 4)
```
oyeb: To ensure only digits: `validate_format(:last_digits, ~r/^\d{4}$/)`, since this counts length too, we don't need the `validate_length`.

apps/snitch_core/lib/core/data/schema/payment/card_payment.ex
```diff
@@ -55,4 +57,14 @@ defmodule Snitch.Data.Schema.CardPayment do
   def changeset(payment, params, :update) do
     cast(payment, params, @update_fields)
   end
+
+  def is_card_id_present(payment) do
```
oyeb: Rename this to `assoc_card`

apps/snitch_core/priv/repo/migrations/20180329105214_create_card_table.exs
```diff
@@ -0,0 +1,18 @@
+defmodule Snitch.Repo.Migrations.CreateCardTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:snitch_cards) do
+      add(:user_id, references(:snitch_users, on_delete: :delete_all), null: false)
+      add(:brand, :string, null: false)
+      add(:month, :integer, null: false)
+      add(:year, :integer, null: false)
+      add(:last_digits, :string, null: false)
```
oyeb: `size: 4`

apps/snitch_core/test/support/factory/factory.ex
```diff
@@ -83,6 +92,16 @@ defmodule Snitch.Factory do
     }
   end
 
+  def credit_card_factory do
```
oyeb: This should be named `card_factory`

apps/snitch_core/test/data/schema/payment/card_payment_test.exs
```diff
@@ -13,18 +26,100 @@ defmodule Snitch.Data.Schema.CardPaymentTest do
...
+          :create
+        )
+
+      %{valid?: validity} = card_payment
+
+      assert validity
+      assert {:ok, _} = Repo.insert(card_payment)
+    end
+
+    test "with already having card_id", context do
```
oyeb: I don't understand the point of this test. Did you want to test if the same card can be used for 2 orders? I don't think that needs to be tested as the database design guarantees that.
The use of `:rand` in this test is wrong, instead you should use the factory methods to insert dummy `orders`, `payments` and `card_payments`.

apps/snitch_core/test/data/model/payment/card_payment_test.exs
```diff
@@ -9,6 +9,7 @@ defmodule Snitch.Data.Model.CardPaymentMethodTest do
   setup :user_with_address
   setup :an_order
   setup :payment_methods
+  setup :card
   setup :card_payment
```
oyeb: Please add a test to check if we can update the `card` used in some `card_payment`

apps/snitch_core/test/data/schema/payment/card_payment_test.exs
```diff
@@ -13,18 +26,100 @@ defmodule Snitch.Data.Schema.CardPaymentTest do
...
+          %CardPayment{},
+          %{payment_id: check.id, card: Map.put(@card, :user_id, user.id)},
+          :create
+        )
+
       assert {:error, %Ecto.Changeset{errors: errors}} = Repo.insert(card_payment)
       assert errors == [payment_id: {"does not refer a card payment", []}]
     end
+
+    test "with all good", context do
```
oyeb: with all good _what_?

apps/snitch_core/test/data/schema/payment/card_payment_test.exs
```diff
@@ -13,18 +26,100 @@ defmodule Snitch.Data.Schema.CardPaymentTest do
...
+      %{user: user, card_method: card_m, order: order} = context
+
+      payment_struct = %Payment{@payment | payment_method_id: card_m.id, order_id: order.id}
```
oyeb: Why do you have to make this struct? Isn't this identical to `context.ccd`?

apps/snitch_core/lib/core/data/schema/payment/card_payment.ex
```diff
@@ -22,13 +22,14 @@ defmodule Snitch.Data.Schema.CardPayment do
...
     belongs_to(:payment, Payment)
+    belongs_to(:card, Card)
 
     timestamps()
   end
 
   @update_fields ~w(response_code response_message avs_response cvv_response)a
   @create_fields [:payment_id | @update_fields]
-
+  @temp ~w(card_id)a
```
oyeb: I think `card_id` should be part of `@update_fields`

apps/snitch_core/lib/core/data/schema/payment/card_payment.ex
```diff
@@ -55,4 +57,14 @@ defmodule Snitch.Data.Schema.CardPayment do
   def changeset(payment, params, :update) do
     cast(payment, params, @update_fields)
   end
+
+  def is_card_id_present(payment) do
+    if payment.params["card_id"] == nil do
+      payment
+      |> cast_assoc(:card, with: &Card.changeset(&1, &2, :create), required: true)
+      |> foreign_key_constraint(:card_id)
```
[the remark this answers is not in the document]
gopalshimpi: ??? why?

apps/snitch_core/lib/core/data/schema/payment/card_payment.ex
```diff
@@ -55,4 +57,14 @@ defmodule Snitch.Data.Schema.CardPayment do
   def changeset(payment, params, :update) do
     cast(payment, params, @update_fields)
   end
+
+  def is_card_id_present(payment) do
+    if payment.params["card_id"] == nil do
```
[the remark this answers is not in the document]
gopalshimpi: No its not atom.
```
%{
  "card" => %{
    brand: "VISA",
    last_digits: "0821",
    month: 12,
    name_on_card: "Harry Potter",
    user_id: 945,
    year: 2050
  },
  "card_id" => 521,
  "payment_id" => 1520
}
```

apps/snitch_core/test/data/schema/payment/card_payment_test.exs
gopalshimpi: I want to test the `assoc_card` function for `card_payment` schema. Because when we are having card_id (card is previously used & saved) then it wont make association again. So if card_id is present the `else` part of `assoc_card` will be executed. To to meet the coverage for the else part I have wrote this test.

apps/snitch_core/test/data/model/payment/card_payment_test.exs
gopalshimpi: Sure

apps/snitch_core/test/data/schema/payment/card_payment_test.exs
gopalshimpi: New card is added to card_payment.

apps/snitch_core/lib/core/data/schema/payment/card_payment.ex
gopalshimpi: yes it can be.

apps/snitch_core/test/data/schema/payment/card_payment_test.exs
gopalshimpi: No it is not identical to context.ccd, because it returning the payment.id (ccd.id) which is already taken.

apps/snitch_core/test/data/schema/payment/card_payment_test.exs
```diff
@@ -53,7 +53,7 @@ defmodule Snitch.Data.Schema.CardPaymentTest do
       assert errors == [payment_id: {"does not refer a card payment", []}]
     end
 
-    test "with all good", context do
+    test "when added new card", context do
```
oyeb: This test reads like this: "CardPayment records when added new card" and this does not make sense. Please use a new describe block and improve the title as well -- it should explain the scenario unambigously.

apps/snitch_core/test/data/schema/payment/card_payment_test.exs
oyeb: You are mistaken, please see [line 135 of `factory.ex`](https://github.com/aviabird/snitch/blob/card_schema_init/apps/snitch_core/test/support/factory/factory.ex#L135) and you'll see that `context.ccd` has the same associations as the struct you are building manually.
Moreover, what is the point of creating a new `card_payment` for this test?
What do _you_ mean by "added new card"? A new card is being added for both the tests above ([L28](https://github.com/aviabird/snitch/blob/card_schema_init/apps/snitch_core/test/data/schema/payment/card_payment_test.exs#L28), [L42](https://github.com/aviabird/snitch/blob/card_schema_init/apps/snitch_core/test/data/schema/payment/card_payment_test.exs#L42))

apps/snitch_core/test/data/schema/payment/card_payment_test.exs
```diff
@@ -13,18 +26,98 @@ defmodule Snitch.Data.Schema.CardPaymentTest do
...
+          %CardPayment{},
+          %{payment_id: check.id, card: Map.put(@card, :user_id, user.id)},
+          :create
+        )
+
       assert {:error, %Ecto.Changeset{errors: errors}} = Repo.insert(card_payment)
       assert errors == [payment_id: {"does not refer a card payment", []}]
     end
+
+    test "when added new card", context do
```
oyeb: This test will not cover the `else` block of `assoc_card`, and this execution flow path is already _covered_ in tests above. IMO, this test is pointless.

apps/snitch_core/test/data/schema/payment/card_payment_test.exs
oyeb: I see that you want to cover the `else` block, but why do you have to _"use a card in a previous payment"_ to test this? All this test requires is that you:
1. create a `card` using its factory method,
2. create a `card_payment` as done on [L109](https://github.com/aviabird/snitch/pull/45/files#diff-f65d50d77a98282f5baf6a439233a14aR109)

apps/snitch_core/lib/core/data/schema/payment/card_payment.ex
```diff
@@ -55,4 +57,15 @@ defmodule Snitch.Data.Schema.CardPayment do
   def changeset(payment, params, :update) do
     cast(payment, params, @update_fields)
   end
+
+  def assoc_card(payment) do
+    if payment.params["card_id"] == nil do
+      payment
+      |> cast_assoc(:card, with: &Card.changeset(&1, &2, :create), required: true)
+      |> foreign_key_constraint(:card_id)
```
oyeb: This can be removed because the `card` is being created in-situ.

apps/snitch_core/lib/core/data/schema/card.ex
```diff
@@ -38,10 +35,10 @@ defmodule Snitch.Data.Schema.Card do
     |> validate_number(
       :month,
       less_than_or_equal_to: 12,
-      greater_than_or_equal_to: @current_month
+      greater_than_or_equal_to: DateTime.utc_now() |> Map.fetch!(:month)
```
oyeb: Though this is correct, a more readable and slightly efficient way if to get the current year and month together at the top,
```elixir
def changeset(card, params, :create) do
  %{year: current_year, month: current_month} = DateTime.utc_now()
  ...
end
```

apps/snitch_core/lib/core/data/schema/payment/card_payment.ex
```diff
@@ -43,8 +43,8 @@ defmodule Snitch.Data.Schema.CardPayment do
 
   def changeset(payment, params, :create) do
     payment
-    |> cast(params, @create_fields ++ @temp)
-    |> is_card_id_present
+    |> cast(params, @create_fields)
+    |> assoc_card
```
oyeb: Please put parens around all function calls.

apps/snitch_core/lib/core/data/schema/card.ex
```diff
@@ -1,47 +1,84 @@
...
+  def build_card(card) do
+    card
+    |> split_name(Map.has_key?(card.changes, :name_on_card))
+    |> extract_last_digits(
+      Map.has_key?(card.changes, :card_number),
+      Map.has_key?(card.changes, :brand)
+    )
+  end
+
+  defp split_name(card, true) do
```
oyeb: This is a lossy function! `split_name("A. K. Jadhav")` will simply throw away the middle `K.`. IMO opinion we should take a simple approach:
1. If `:name_on_card` is set, we will leave `:first_name` and `:last_name` empty.
2. If `:name_on_card` is empty and `:first_name`, `:last_name` are set, we use them to build `:name_on_card`.
3. If all name fields are empty, we mark changeset invalid.

apps/snitch_core/lib/core/data/schema/card.ex
```diff
@@ -1,47 +1,84 @@
...
+
+  def changeset(card, params, :update) do
+    cast(card, params, @update_fields)
+  end
+
+  def build_card(card) do
+    card
+    |> split_name(Map.has_key?(card.changes, :name_on_card))
+    |> extract_last_digits(
+      Map.has_key?(card.changes, :card_number),
```
oyeb: All these checks can be inside the body of `extract_last_digit` _or in the guard clause of `extract_last_digit`.

apps/snitch_core/lib/core/data/schema/payment/card_payment.ex
```diff
@@ -22,11 +22,12 @@ defmodule Snitch.Data.Schema.CardPayment do
     field(:cvv_response, :string)
 
     belongs_to(:payment, Payment)
+    belongs_to(:card, Card)
 
     timestamps()
   end
 
-  @update_fields ~w(response_code response_message avs_response cvv_response)a
+  @update_fields ~w(response_code response_message avs_response cvv_response card_id)a
```
oyeb: `card_id` should not be updatable.

apps/snitch_core/lib/core/data/schema/card.ex
```diff
@@ -1,47 +1,84 @@
...
   def changeset(card, params, action)
 
   def changeset(card, params, :create) do
     %{year: current_year, month: current_month} = DateTime.utc_now()
 
+    # for now only for VISA cards.
+
     card
-    |> cast(params, @required_fields ++ @optional_fields)
+    |> cast(params, @required_fields)
```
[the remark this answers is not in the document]
gopalshimpi: It should be casted ? It is virtual field it is not in db.

apps/snitch_core/test/data/model/payment/card_payment_test.exs
gopalshimpi: No need for this.

PR #55, opened 2018-04-30
Add selling_price to Variant, and tests

arjun289:
* Modified `discontinue_on` type from `naive_datetime` 
to `utc_datetime`, added validation.
* Added tests for the variant schema.
* Update `line_item` and `payment` tests.

### DIFF
--- a/apps/snitch_core/lib/core/data/schema/payment/payment.ex
+++ b/apps/snitch_core/lib/core/data/schema/payment/payment.ex
@@ -35,7 +35,7 @@ defmodule Snitch.Data.Schema.Payment do
     payment
     |> cast(params, @create_fields)
     |> validate_required(@create_fields)
-    |> validate_discriminator(:payment_type, @payment_types)
+    |> validate_inclusion(:payment_type, @payment_types)
     |> foreign_key_constraint(:payment_method_id)
     |> common_changeset()
   end
@@ -62,30 +62,4 @@ defmodule Snitch.Data.Schema.Payment do
     |> validate_amount(:amount)
     |> unique_constraint(:slug)
   end
-
-  @spec validate_discriminator(Ecto.Changeset.t(), atom, list) :: Ecto.Changeset.t()
-  defp validate_discriminator(%Ecto.Changeset{valid?: true} = changeset, key, permitted) do
-    {_, discriminator} = fetch_field(changeset, key)
-
-    if discriminator in permitted do
-      changeset
-    else
-      add_error(changeset, :payment_type, "'#{discriminator}' is invalid", validation: :inclusion)
-    end
-  end
-
-  defp validate_discriminator(changeset, _, _), do: changeset
-
-  @spec validate_amount(Ecto.Changeset.t(), atom) :: Ecto.Changeset.t()
-  defp validate_amount(%Ecto.Changeset{valid?: true} = changeset, key) do
-    {_, amount} = fetch_field(changeset, key)
-
-    if Decimal.cmp(amount.amount, Decimal.new(0)) != :lt do
-      changeset
-    else
-      add_error(changeset, :amount, "must be greater than 0", validation: :amount)
-    end
-  end
-
-  defp validate_amount(changeset, _), do: changeset
 end
--- a/apps/snitch_core/lib/core/data/schema/variant.ex
+++ b/apps/snitch_core/lib/core/data/schema/variant.ex
@@ -4,60 +4,58 @@ defmodule Snitch.Data.Schema.Variant do
   """
 
   use Snitch.Data.Schema
-
   import Ecto.Query
 
   alias Snitch.Data.Schema.StockItem
-  alias Snitch.Repo
   alias Money.Ecto.Composite.Type, as: MoneyType
+  alias Snitch.Repo
 
   @type t :: %__MODULE__{}
 
   schema "snitch_variants" do
-    field(:sku, :string, default: "")
+    field(:sku, :string)
     field(:weight, :decimal, default: Decimal.new(0))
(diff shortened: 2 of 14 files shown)


apps/snitch_core/test/support/factory/factory.ex
```diff
@@ -175,4 +175,14 @@ defmodule Snitch.Factory do
     card_count = Map.get(context, :card_count, 1)
     [cards: insert_list(card_count, :card, user_id: user.id)]
   end
+
+  def get_date(offset_days) do
```
oyeb: Better signature: `offset_date_by(DateTime.t(), offset_days :: integer) :: DateTime.t()`

apps/snitch_core/lib/core/data/schema/payment/payment.ex
```diff
@@ -75,17 +75,4 @@ defmodule Snitch.Data.Schema.Payment do
   end
 
   defp validate_discriminator(changeset, _, _), do: changeset
```
SagarKarwande: Do we need these arguments `validate_discriminator(:payment_type, @payment_types)`
As this function only validate payment types and cannot be reused as it specific for one thing.
This arguments can be removed and use them directly in the function definition iteself.
**OR** 
Ecto has similar thing for it:
`validate_inclusion(changeset, :payment_type, ["chk", "ccd"])`

apps/snitch_core/lib/core/data/schema/variant.ex
```diff
@@ -4,60 +4,58 @@ defmodule Snitch.Data.Schema.Variant do
...
 
     has_many(:stock_items, StockItem)
 
     timestamps()
   end
 
-  @permitted_fields ~w(sku weight height width depth is_master)a ++
-                      ~w(cost_price position track_inventory discontinue_on)a
+  @cast_fields ~w(sku weight height width depth selling_price)a ++
+                 ~w(cost_price position track_inventory discontinue_on)a
```
SagarKarwande: Can this be written as
`@cast_fields ~w(sku weight height width depth is_master cost_price position track_inventory discontinue_on)a`

apps/snitch_core/lib/core/tools/validations.ex
```diff
@@ -0,0 +1,59 @@
...
+  """
+
+  import Ecto.Changeset
+
+  @doc """
+  Validates that the amount (of type `Money.t`) under the `key` in `changeset`
+  is non-negative.
+  """
+  @spec validate_amount(Ecto.Changeset.t(), atom) :: Ecto.Changeset.t()
+  def validate_amount(%Ecto.Changeset{valid?: true} = changeset, key) when is_atom(key) do
```
SagarKarwande: This can be removed as Ecto has same method for it `validate_number(changeset, field, opts)`
Ecto doc link :  [validate_number](https://hexdocs.pm/ecto/Ecto.Changeset.html#validate_number/3)
Check the defination : [validate_number](https://github.com/elixir-ecto/ecto/blob/v2.2.10/lib/ecto/changeset.ex#L1726)

apps/snitch_core/lib/core/data/schema/variant.ex
```diff
@@ -4,60 +4,58 @@ defmodule Snitch.Data.Schema.Variant do
...
+  def create_changeset(%__MODULE__{} = variant, params) do
     variant
-    |> cast(attrs, @permitted_fields)
+    |> cast(params, @cast_fields)
+    |> validate_required(@required_fields)
     |> unique_constraint(:sku)
-
-    # Ensures a new variant takes the product master price when price is not supplied
-    # Ensure variants? are not soft deleted
+    |> validate_amount(:selling_price)
```
SagarKarwande: Use `validate_number(changeset, field, opts)` instead

apps/snitch_core/lib/core/tools/validations.ex
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def validate_amount(changeset, _), do: changeset
+
+  @doc """
+  Validates that the given date (of type `DateTime.t`) under the `key` in
+  `changeset` is in the future wrt. `DateTime.utc_now/0`.
+  """
+  @spec validate_future_date(Ecto.Changeset.t(), atom) :: Ecto.Changeset.t()
+  def validate_future_date(changeset, key) when is_atom(key) do
```
SagarKarwande: Is there a need to pattern match changset `is_valid` here ?

apps/snitch_core/lib/core/tools/validations.ex
arjun289: we can not use this one here, amount here is not a number but a `MoneyType`.

apps/snitch_core/lib/core/tools/validations.ex
arjun289: yeah missed this updating.

apps/snitch_core/lib/core/data/schema/payment/payment.ex
oyeb: `validate_inclusion` only validates the `changes` but not the data of the changeset. All changeset functions assume that the `data` is always valid.
If I do the following, `validate_inclusion` will not complain whereas `validate_discriminator` will (use of `fetch_field`)
```elixir
Schema.Payment.create_changeset(
    %Schema.Payment{payment_type: "foo"},
    %{
      amount: Money.zero(:USD),
      payment_method_id: 2,
      slug: "foo",
      order_id: 3
    }
)
```

Generally, ecto validators work on `changes`, and I went paranoid in this case and tried to protect this extremely unlikely case.
```

PR #56, opened 2018-05-01
Add user, address, variant, order, lineitem seed data

oyeb:
**Used only `insert_all` for inserts to guarantee speed and handle constraint conflicts.**

* There's some randomness, just for giggles :hibiscus:
* Oops, order slug had no unique constraint, fixed that.

[Delivers #157229687](https://www.pivotaltracker.com/story/show/157229687).

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -3,7 +3,7 @@
     "apps/*/{lib,config}/**/*.{ex,exs}", # lib and config
     "apps/*/test/**/*.{ex,exs}", # tests
     # "apps/*/priv/repo/migrations/*.{ex,exs}", # migrations
-    "apps/*/priv/repo/*.{ex,exs}", # seeds
+    "apps/*/priv/repo/seed/*.{ex,exs}", # seeds
     "apps/*/mix.exs", # mix files
     "mix.exs", # top-level
   ]
--- a/apps/snitch_core/priv/repo/seed/orders.ex
+++ b/apps/snitch_core/priv/repo/seed/orders.ex
@@ -1,0 +1,153 @@
+defmodule Snitch.Seed.Orders do
+  @moduledoc false
+
+  alias Ecto.DateTime
+  alias Snitch.Repo
+  alias Snitch.Data.Schema.{Variant, Order, Address, User, LineItem}
+
+  @line_item %{
+    variant_id: nil,
+    order_id: nil,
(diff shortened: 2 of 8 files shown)


.formatter.exs
```diff
@@ -3,7 +3,7 @@
     "apps/*/{lib,config}/**/*.{ex,exs}", # lib and config
     "apps/*/test/**/*.{ex,exs}", # tests
     # "apps/*/priv/repo/migrations/*.{ex,exs}", # migrations
```
SagarKarwande: Why are we skipping formatting for migrations? :confused:

apps/snitch_core/priv/repo/seed/orders.ex
```diff
@@ -0,0 +1,151 @@
...
+
+  def variants do
+    0
+    |> Stream.iterate(&(&1 + 1))
+    |> Stream.map(&"shoes-nike-#{&1}")
+    |> Stream.map(fn sku -> %{random_variant() | sku: sku} end)
+  end
+
+  def random_variant do
+    %{
```
SagarKarwande: Why are we not using `%Variant{}`

apps/snitch_core/priv/repo/seed/orders.ex
```diff
@@ -0,0 +1,151 @@
...
+    variant_id: nil,
+    order_id: nil,
+    quantity: nil,
+    unit_price: nil,
+    total: nil,
+    inserted_at: DateTime.utc(),
+    updated_at: DateTime.utc()
+  }
+
+  @order %{
```
SagarKarwande: Why we are not using `%Order{}`

apps/snitch_core/priv/repo/seed/orders.ex
oyeb: Because [`Ecto.Repo.insert_all`](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert_all/3) requires a `[map] | keyword`.

.formatter.exs
oyeb: No real reason, we can turn that on if you wish. We arbitrarily decided to turn it off when we started.
But, migrations look a bit cleaner without the parens don't you think? :stuck_out_tongue: 
Let's vote: those in favour react with a :+1:, those against with :-1:

PR #59, opened 2018-05-03
Taxonomy

SagarKarwande:
* Add taxonomy domain to create taxonomy easily.
* Adds taxonomy helper to create taxonomy with ease.
* Test and Docs for all module.
* Seed for taxonomy

### DIFF
--- a/apps/snitch_core/lib/core/data/schema/taxanomy/taxon.ex
+++ b/apps/snitch_core/lib/core/data/schema/taxanomy/taxon.ex
@@ -1,0 +1,28 @@
+defmodule Snitch.Data.Schema.Taxon do
+  @moduledoc """
+
+  """
+  use Snitch.Data.Schema
+  use AsNestedSet, scope: [:taxonomy_id]
+
+  alias Snitch.Data.Schema.{Taxonomy, Taxon}
+
+  @type t :: %__MODULE__{}
+
+  schema "snitch_taxons" do
+    field(:name, :string)
+    field(:lft, :integer)
+    field(:rgt, :integer)
+
+    belongs_to(:taxonomy, Taxonomy)
+    belongs_to(:parent, Taxon)
+    timestamps()
+  end
+
+  @cast_fields ~w(name parent_id taxonomy_id lft rgt)
+
+  def changeset(taxon, params) do
+    taxon
+    |> cast(params, @cast_fields)
+  end
+end
--- a/apps/snitch_core/lib/core/domain/taxonomy/taxonomy.ex
+++ b/apps/snitch_core/lib/core/domain/taxonomy/taxonomy.ex
@@ -1,0 +1,82 @@
+defmodule Snitch.Domain.Taxonomy do
+  @moduledoc """
+  Interface for handling Taxonomy. It provides functions to modify Taxonomy.
+  """
+
+  use Snitch.Domain
+
+  import AsNestedSet.Modifiable
+  import AsNestedSet.Queriable, only: [dump_one: 2]
+
+  alias Snitch.Data.Schema.Taxon
+
+  @doc """
+    Adds child taxon to left, right or child of parent taxon.
+
+    Positon can take follwoing values.
+    Position - :left | :right | :child
+  """
+  @spec add_taxon(Taxon.t(), Taxon.t(), atom) :: Taxon.t()
+  def add_taxon(%Taxon{} = parent, %Taxon{} = child, position) do
+    child =
+      %Taxon{child | taxonomy_id: parent.taxonomy.id}
+      |> Repo.preload(:taxonomy)
+
+    child
+    |> create(parent, position)
+    |> AsNestedSet.execute(Repo)
+  end
+
+  @doc """
+  Adds taxon as root to the taxonomy
+  """
+  @spec add_root(Taxon.t()) :: Taxon.t()
+  def add_root(%Taxon{} = root) do
+    root
+    |> create(:root)
+    |> AsNestedSet.execute(Repo)
+  end
+
+  @doc """
+  Get the root for the taxonomy of passed the taxon
+  """
+  @spec get_root(Taxon.t()) :: Taxon.t()
+  def get_root(%Taxon{} = taxon) do
+    AsNestedSet.root(Taxon, %{taxonomy_id: taxon.taxonomy_id})
+    |> AsNestedSet.execute(Repo)
+  end
+
+  @doc """
+  Traverse in-order the taxonomy tree and return the tuple of root and list of
+  traversed taxons
+  """
+  @spec inorder_list(Taxon.t()) :: {Taxon.t(), [Taxon.t()]}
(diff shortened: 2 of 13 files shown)


apps/snitch_core/lib/core/domain/taxonomy/taxonomy.ex
```diff
@@ -0,0 +1,81 @@
...
+      fn node, acc -> {node, acc} end
+    )
+    |> AsNestedSet.execute(Repo)
+  end
+
+  @doc """
+  Dumps the taxonomy in tuple form.
+  """
+  @spec dump_taxonomy(Taxon.t()) :: {Taxon.t(), [{Taxon.t(), []}]}
+  def dump_taxonomy(%Taxon{} = taxon) do
```
oyeb: Does this dump the complete taxonomy, or the tree rooted at `taxon`?

apps/snitch_core/lib/core/data/schema/taxanomy/taxon.ex
```diff
@@ -0,0 +1,28 @@
...
+    field(:rgt, :integer)
+
+    belongs_to(:taxonomy, Taxonomy)
+    belongs_to(:parent, Taxon)
+    timestamps()
+  end
+
+  @cast_fields ~w(name parent_id taxonomy_id lft rgt)
+
+  def changeset(taxon, params) do
```
oyeb: This does not seem to be used.

apps/snitch_core/lib/core/tools/helpers/taxonomy.ex
```diff
@@ -0,0 +1,59 @@
...
+            {"Mugs", []},
+            {"Clothing",
+             [
+               {"Shirts", []},
+               {"T-Shirts", []}
+          ]}
+        ]}
+  """
+  @spec create_taxonomy({String.t(), []}) :: Taxonomy.t()
+  def create_taxonomy({parent, children}) do
```
oyeb: 1. Why is the argument a tuple?
2. This function and `create_taxon/2` should make use of `Ecto.Multi`.
3. This can be sped up a bit by using `Repo.insert_all` for each level of the tree, but the current approach is good enough.

apps/snitch_core/test/domain/taxonomy/taxonomy_test.exs
```diff
@@ -0,0 +1,232 @@
...
+  use ExUnit.Case, async: true
+  use Snitch.DataCase
+
+  import Snitch.Factory
+
+  alias FakerElixir, as: Faker
+  alias Snitch.Domain.Taxonomy
+  alias Snitch.Data.Schema.Taxon
+
+  # Will create a better code for this to create taxonomy from list or map
```
oyeb: The current approach is good enough, I still think of this as a temporary implementation of taxonomy so I wouldn't spend any time in perfecting it.
1. **But this test can still be sped up!**
    -  Because we are starting off with a relatively large taxonomy, we can do all the 4 addition variations in just one test and match the result with what we expect.
    - Another reason to merge the cases is that there's no point in testing the behavior of `as_nested_set`. We have just one `add_taxon` that _does not care about the `position` argument!_
2. Could you please format the diagram below as the n-ary tree? I cannot do an inorder traversal of this diagram in my mind.

apps/snitch_core/test/domain/taxonomy/taxonomy_test.exs
```diff
@@ -0,0 +1,232 @@
...
+      taxon = %Taxon{name: "root", taxonomy_id: taxonomy.id}
+
+      root = Taxonomy.add_root(taxon)
+
+      assert root.name == taxon.name
+      assert root.taxonomy_id == taxonomy.id
+    end
+
+    test "create root for invalid taxonomy" do
+      taxon = %Taxon{name: "root", taxonomy_id: Faker.Number.digits()}
```
oyeb: This seems like the only use of Faker in the whole PR. I would have just put `-1` to get the same behaviour :smile: 
That said, this test is pointless, as it tests Ecto's behaviour.

apps/snitch_core/test/domain/taxonomy/taxonomy_test.exs
```diff
@@ -0,0 +1,232 @@
...
+               "Kitchen & Tables",
+               "Table Covers",
+               "Mat & Napkins",
+               "Home Decor"
+             ]
+    end
+  end
+
+  describe "add_root/1" do
+    test "create root for empty taxonomy" do
```
oyeb: This would work, but can I create a root for a non-empty taxonomy?

apps/snitch_core/test/domain/taxonomy/taxonomy_test.exs
```diff
@@ -0,0 +1,232 @@
...
+          {table_cover, []},
+          {mat_napkins, []}
+        ]},
+       {home_decor, []}
+     ]}
+  end
+
+  describe "add_taxon/2" do
+    test "add child taxon" do
+      {_, [{target, _} | _]} = create_taxonomy()
```
oyeb: Hard to see which taxon `target` is, could you assert its name to make this more readable?

apps/snitch_core/test/domain/taxonomy/taxonomy_test.exs
```diff
@@ -0,0 +1,232 @@
...
+      child = %Taxon{name: "Carpet"}
+      taxon = Taxonomy.add_taxon(target, child, :child)
+
+      taxonomies = get_taxons_name_list(target)
+
+      assert taxon.name == "Carpet"
+      assert taxon.taxonomy_id == target.taxonomy_id
+      assert taxon.parent_id == target.id
+
+      assert taxonomies == [
```
oyeb: Technically, an inorder traversal does not uniquely determine the tree structure (two different traversals are needed) It's best to match on the n-ary tuple (of just names, perhaps).

apps/snitch_core/test/support/factory/taxonomy.ex
```diff
@@ -0,0 +1,22 @@
...
+
+      def taxonomy_factory() do
+        %Taxonomy{
+          name: sequence("Taxonomy")
+        }
+      end
+
+      def taxon_factory() do
+        %Taxon{
+          name: sequence("Taxonomy")
```
oyeb: typo

apps/snitch_core/test/tools/helpers/taxonomy_test.exs
```diff
@@ -0,0 +1,50 @@
...
+      taxonomy =
+        TaxonomyHelper.create_taxonomy({"Root", []})
+        |> Repo.preload(:root)
+
+      taxons = get_taxons_name_list(taxonomy.root)
+      assert taxons == ["Root"]
+    end
+  end
+
+  defp get_taxons_name_list(taxon) do
```
oyeb: This function can move to the helper module as well. Though the utility of the flat list of names is rather limited.

PR #70, opened 2018-05-13
Shipment packer sans shipping-methods and zones

oyeb:
The [default package builder story](https://www.pivotaltracker.com/story/show/157490475).

Gives default packages, but
* Could be "restrictive" for some store owners who'd like to split
backorder-only items into costlier shipments automatically.
* The items of a package are guaranteed to have the same `ShippingCategory`.

[delivers #157490475]


apps/snitch_core/lib/core/domain/shipment.ex
```diff
@@ -0,0 +1,117 @@
+defmodule Snitch.Domain.Shipment do
+  @moduledoc """
+  The coordinator, packer, estimator and prioritizer -- all in one.
+
+  The package struct:
+  ```
+  %{
+    hash: TBD
+    items: [
```
arjun289: closing `]` missing.

apps/snitch_core/lib/core/domain/shipment.ex
```diff
@@ -0,0 +1,117 @@
...
+    # |> Stream.map(&attach_zone_info(order.shipping_address))
+    |> Enum.map(&split_by_category/1)
+    |> List.flatten()
+  end
+
+  defp package(stock_location, line_items) do
+    # TODO: Bad price!
+    #
+    # Should it be (on_hand * selling_price) or (quantity * selling_price)?
+    # Currently it is (quantity * selling_price)
```
arjun289: little confused here, where are you setting the `price`?

apps/snitch_core/test/support/order_case.ex
```diff
@@ -0,0 +1,60 @@
...
+    order_id: nil,
+    inserted_at: Ecto.DateTime.utc(),
+    updated_at: Ecto.DateTime.utc()
+  }
+
+  @doc """
+  Returns a list of line_item `map`s after zipping `variants` and
+  `quanities`.
+
+  The price fields are not computed. This result is suitable for a
```
arjun289: please put a note that the number of variants should be same as quantity.

oyeb: It need not be, and that's desirable. The zipping will stop when one of the enumerables is exhausted. A handy way of controlling the number of line_items -- just make the quantity list small :smile:

PR #72, opened 2018-05-14
Add TaxCalculator and related functionality

arjun289:
The PR covers the following
#### TaxRate [ #157502102](https://www.pivotaltracker.com/story/show/157502102)
* schema
   - added schema and tests
   - added migration
* model
   - added CRUD functions
   - added tests

#### UserConfig [#157502146](https://www.pivotaltracker.com/story/show/157502146)
* Added Userconfig module.
  Added tests

#### Calculator  [#157147786](https://www.pivotaltracker.com/story/show/157147786)
* Calculator behavior
  - module functions.
   - tests.
* DefaultTaxCalculator
   - added default calculator module
   - added tests


apps/snitch_core/priv/repo/migrations/20180509123728_add_tax_rate_table.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Snitch.Repo.Migrations.AddTaxRateTable do
+  use Ecto.Migration
+
+  def change do
+    create  table(:snitch_tax_rates) do
+      add :name, :string
+      add :value, :decimal
+      add :calculator, :string
```
oyeb: Please put some NOT NULL constraints here. IMO the name should be unique too.

apps/snitch_core/test/calculator/calculator_test.exs
```diff
@@ -0,0 +1,32 @@
+defmodule Snitch.CalculatorTest do
+  use ExUnit.Case, async: false
```
oyeb: `async: true` pls!

apps/snitch_core/test/data/schema/tax_rate_test.exs
```diff
@@ -0,0 +1,56 @@
...
+    included_in_price: false,
+    calculator: :DefaultCalculator
+  }
+
+  setup :tax_rate_params
+
+  describe "creation" do
+    test "with valid params", context do
+      %{tax_rate_params: params} = context
+      changeset = %{valid?: validity} = TaxRate.create_changeset(%TaxRate{}, params)
```
oyeb: You should assert on the modified calculator string as well.

apps/snitch_core/test/support/factory/factory.ex
```diff
@@ -201,4 +211,17 @@ defmodule Snitch.Factory do
     count = Map.get(context, :tax_category_count, 3)
     [tax_categories: insert_list(count, :tax_category)]
   end
+
+  def tax_rate(_context) do
```
oyeb: You can emulate `tax_rate/1` by using `tax_rates/1`. `tax_rate/1` does not use it's argument.
**Ideally, you should pick the `zone` from `context` so that you minimise zone and tax category insertions in the test setup.**

_Moreover, `def tax_rate_factory` injects a `tax_rate/0` anyways._

apps/snitch_core/test/tools/user_config_test.exs
```diff
@@ -0,0 +1,35 @@
...
+  describe "configured properly" do
+    setup do
+      Application.put_env(:snitch_core, :calculators, [TaxCalculator, FlatRateCalculator])
+
+      on_exit(fn ->
+        Application.delete_env(:snitch_core, :calculators)
+      end)
+    end
+
+    test "fetch all calculators" do
```
oyeb: Just a stylistic comment:
This module has nothing to do with calculators. To convey the generality of this module in this test you should put things like `:foo` or `:bar` in the app env instead of `:calculators`.

apps/snitch_core/lib/core/calculator/calculator.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Snitch.Calculator do
```
oyeb: Why is this not inside the `Snitch.Domain` namespace?

apps/snitch_core/test/data/model/tax_rate_test.exs
```diff
@@ -0,0 +1,147 @@
...
+    calculator: "DefaultCalculator"
+  }
+
+  describe "create/1" do
+    test "tax rate successfully" do
+      params = create_tax_params()
+      assert {:ok, _} = TaxRateModel.create(params)
+    end
+
+    test "fails, name missing" do
```
oyeb: You can group all the cast and validate required tests in one test. Pass an empty `params` and all validations will throw errors.

apps/snitch_core/lib/core/data/model/tax_rate.ex
```diff
@@ -0,0 +1,100 @@
...
+
+  @doc """
+  Creates a TaxRate with supplied `params`.
+  """
+  @spec create(map) :: {:ok, TaxRate.t()} | {:error, Ecto.Changeset.t()}
+  def create(params) do
+    params =
+      if Map.has_key?(params, :calculator) do
+        %{calculator: calc} = params
+        %{params | calculator: String.to_atom(calc)}
```
oyeb: Since the schema declares `:calculator` to be an `atom`, why would anyone pass a `string` here? In fact, you should never convert user-supplied strings into atoms! Passing a string should be strictly forbidden.

apps/snitch_core/test/data/model/tax_rate_test.exs
```diff
@@ -0,0 +1,147 @@
...
+
+    test "successfuly calculator field", context do
+      params = %{calculator: "FlatRateCalculator"}
+      %{tax_rate: tr} = context
+      assert {:ok, tr_new} = TaxRateModel.update(params, tr)
+      assert tr_new.id == tr.id
+      refute tr_new.calculator == tr.calculator
+    end
+
+    test "successful without instance", context do
```
oyeb: Please don't test the behaviour of the `QueryHelper` here! You just have to cover one of the clauses.

apps/snitch_core/test/data/model/tax_rate_test.exs
```diff
@@ -0,0 +1,147 @@
...
+      %{tax_rate: tr} = context
+      params = %{name: "Germany", id: tr.id}
+      assert {:ok, tr_new} = TaxRateModel.update(params)
+      assert tr_new.id == tr.id
+      refute tr_new.name == tr.name
+    end
+
+    test "failed, name missing", context do
```
oyeb: Why should this fail? What if all I want to do is update the `value`?

apps/snitch_core/test/data/model/tax_rate_test.exs
```diff
@@ -0,0 +1,147 @@
...
+      params = %{name: "", id: tr.id}
+      assert {:error, cs} = TaxRateModel.update(params)
+      assert %{name: ["can't be blank"]} = errors_on(cs)
+    end
+  end
+
+  describe "delete/1" do
+    setup :tax_rate
+
+    test "tax category successfully, with instance", context do
```
oyeb: 1. typo
2. Please don't test the behaviour of the `QueryHelper` here. Just test one of the clauses!

apps/snitch_core/test/data/model/tax_rate_test.exs
```diff
@@ -0,0 +1,147 @@
...
+
+    test "no tax rate, is deleted", context do
+      %{tax_rate: tr} = context
+      assert {:ok, tr} = TaxRateModel.delete(tr)
+      refute is_nil(tr.deleted_at)
+      tr_ret = TaxRateModel.get(tr.id)
+      assert is_nil(tr_ret)
+    end
+
+    test "tax category, is deleted", context do
```
oyeb: Too many typos like this one.

apps/snitch_core/test/data/model/tax_rate_test.exs
```diff
@@ -0,0 +1,147 @@
...
+      %{tax_rate: tr} = context
+      {:ok, tr} = TaxRateModel.delete(tr)
+      refute is_nil(tr.deleted_at)
+      tr_ret = TaxRateModel.get(tr.id, false)
+      refute is_nil(tr_ret)
+    end
+  end
+
+  describe "get_all/1" do
```
oyeb: All these asserts can be made in just one test case, where one of the tax rates has been soft deleted.

apps/snitch_core/test/data/model/tax_rate_test.exs
```diff
@@ -0,0 +1,147 @@
...
+    end
+
+    test "tax categories for no record" do
+      Repo.delete_all(TaxRate)
+      tax_rates = TaxRateModel.get_all()
+      assert tax_rates == []
+    end
+  end
+
+  defp create_tax_params do
```
oyeb: This should be _the_ `setup`:
```elixir
setup do
  zone = insert(:zone, zone_type: "S") # no need to wrap it in a %{}
  tax_category = insert(:tax_category)
  params = 
    @valid_params
     |> Map.put(:tax_category_id, tax_category.id)
     |> Map.put(:zone_id, zone.id)
  [zone: zone, tax_category: tax_category, tax_rate_params: params]
end

File: apps/snitch_core/lib/core/data/model/tax_rate.ex [hunk 15]
```diff
@@ -0,0 +1,100 @@
...
+
+  Takes as input 'id' field and an `active` flag.
+  When `active` is false, will return a TaxRate even
+  if it's _soft deleted_.
+
+  > Note, By default tax rate which is present in the table
+  and is __not soft deleted__ is returned.
+  """
+  @spec get(integer, boolean) :: TaxRate.t() | nil
+  def get(id, active \\ true) do
```
SagarKarwande: Can this be written in two separate clauses?

apps/snitch_core/lib/core/data/model/tax_rate.ex
```diff
@@ -0,0 +1,100 @@
...
+  return all the tax_rates, including those which are
+  _soft deleted_.
+
+  > Note the function returns only those tax rates
+    which are not soft deleted by default or if `active` is
+    set to true.
+  """
+
+  @spec get_all(boolean) :: [TaxRate.t()]
+  def get_all(active \\ true) do
```
SagarKarwande: Can this be written in two separate clauses?

apps/snitch_core/lib/core/data/schema/tax_rate.ex
```diff
@@ -0,0 +1,63 @@
+defmodule Snitch.Data.Schema.TaxRate do
+  @moduledoc """
+  Models a TaxRate
+  """
+
+  use Snitch.Data.Schema
+  alias Snitch.Data.Schema.TaxCategory
```
SagarKarwande: Group these aliases `alias Snitch.Data.Schema.{TaxCategory, Zone}`

apps/snitch_core/lib/core/calculator/default_tax_calculator.ex
```diff
@@ -0,0 +1,20 @@
...
+  def compute(tax_rate, %LineItem{} = lineitem) do
+    {:ok, value} =
+      if tax_rate.included_in_price do
+        {:ok, offset} = Money.div(lineitem.total, 1 + tax_rate.value)
+        Money.sub(lineitem.total, offset)
+      else
+        Money.mult(lineitem.total, tax_rate.value)
+      end
+
+    Money.round(value)
```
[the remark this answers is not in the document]
arjun289: this is a matter of contention was going through [this](https://money.stackexchange.com/questions/15051/sales-tax-rounded-then-totaled-or-totaled-then-rounded?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)  but in the end I went with what amazon is following and they round off taxes at lineitems. Open to arguments though.

apps/snitch_core/lib/core/data/schema/tax_rate.ex
```diff
@@ -0,0 +1,63 @@
...
+
+  @type t :: %__MODULE__{}
+
+  schema "snitch_tax_rates" do
+    # associations
+    belongs_to(:tax_category, TaxCategory)
+    belongs_to(:zone, Zone)
+
+    field(:name, :string)
+    field(:value, :decimal)
```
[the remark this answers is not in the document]
arjun289: I haven't done this coz the value field is going to be json as we had discussed. So, no docs with regard to that.

apps/snitch_core/lib/core/calculator/calculator.ex
arjun289: Why should this fall under `domain`? Shouldn't, different services have different namespacing?

apps/snitch_core/lib/core/calculator/default_tax_calculator.ex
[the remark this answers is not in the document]
oyeb: How do you know when they round taxes? What we see is HTML, the internal structs might have insane precision. 
There is (potential) loss of money in summing after rounding using the [half-even](https://en.wikipedia.org/wiki/Rounding#Round_half_to_even) method -- which is the default rounding method of [`Money.round/2`](https://hexdocs.pm/ex_money/Money.html#round/2). If you [round up](https://en.wikipedia.org/wiki/Rounding#Round_half_up) all the time, I don't mind. Still, you should leave the rounding to
1. whoever is displaying the amount.
2. whoever sums them up.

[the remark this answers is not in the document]
arjun289: ohkk, I put in some more research regarding this and realized if the price is very low then this thing will pose a problem when tax value is rounded up.

apps/snitch_core/priv/repo/migrations/20180509123728_add_tax_rate_table.exs
arjun289: `name` can't be unique as two different zones can have a tax with the same name e.g. GST. It's used in both Europe as well as India.

apps/snitch_core/lib/core/data/model/tax_rate.ex
```diff
@@ -0,0 +1,100 @@
...
+  """
+
+  use Snitch.Data.Model
+  alias Snitch.Data.Schema.TaxRate
+
+  @doc """
+  Creates a TaxRate with supplied `params`.
+  """
+  @spec create(map) :: {:ok, TaxRate.t()} | {:error, Ecto.Changeset.t()}
+  def create(params) do
```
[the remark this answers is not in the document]
arjun289: this won't be required, the condition was present to check if `calculator` field is present then convert it to `atom` but it's not being handled here.

apps/snitch_core/priv/repo/migrations/20180509123728_add_tax_rate_table.exs
oyeb: So what's the (compound) primary key for this relation?

apps/snitch_core/lib/core/data/model/tax_rate.ex
arjun289: If you see the use case carefully it's not required.

apps/snitch_core/test/data/model/tax_rate_test.exs
```diff
@@ -38,12 +42,10 @@ defmodule Snitch.Data.Model.TaxRateTest do
...
 
-    test "successfuly calculator field", context do
-      params = %{calculator: "FlatRateCalculator"}
+    test "failed invalid calculator", context do
+      params = %{calculator: BinaryCalculator}
       %{tax_rate: tr} = context
-      assert {:ok, tr_new} = TaxRateModel.update(params, tr)
-      assert tr_new.id == tr.id
-      refute tr_new.calculator == tr.calculator
+      assert {:error, cs} = TaxRateModel.update(params, tr)
```
oyeb: 1. Could you assert on the error message?
2. Is this `true` `tr.calculator == Snitch.Calculator.DefaultCalculator`? Or is this `tr.calculator == Elixir.Snitch.Calculator.DefaultCalculator`?

arjun289: `tr.calculator == Elixir.Snitch.Calculator.DefaultCalculator` because the module names are stored as such.

PR #76, opened 2018-05-21
Weight splitter

SagarKarwande:


apps/snitch_core/test/domain/splitters/weight_test.exs
```diff
@@ -0,0 +1,222 @@
...
+      line_items = line_items_with_price(vs, [1, 1, 1])
+      order = %Order{id: 42, line_items: line_items}
+
+      splitted_packages =
+        order
+        |> Shipment.default_packages()
+        |> WeightSplitter.split()
+
+      # item = {delta quantity}
+      packages_to_assert = [
```
oyeb: This is not a change request, because I don't know of a better way myself :disappointed: 

These tests are a bit fragile because the "order" of packages is assumed. In the _very_ unlikely future, the order could change, and when that happens, these tests will have to be updated.
Helpers (reflection functions) like this [one](https://github.com/aviabird/snitch/blob/package-shipping-method-157490638/apps/snitch_core/test/domain/shipment_test.exs#L135) are not very useful here.

apps/snitch_core/test/domain/splitters/weight_test.exs
```diff
@@ -0,0 +1,222 @@
...
+      |> Enum.map(&assert_package/1)
+    end
+  end
+
+  @tag weights: [50.0, 10.0, 10.0]
+  test "package with more than thresold", %{variants: vs} do
+    line_items = line_items_with_price(vs, [3, 1, 1])
+    order = %Order{id: 42, line_items: line_items}
+
+    splitted_packages =
```
oyeb: typo: splitted is no word. Funny language, english is.

apps/snitch_core/test/domain/splitters/weight_test.exs
```diff
@@ -0,0 +1,222 @@
...
+      order
+      |> Shipment.default_packages()
+      |> WeightSplitter.split()
+
+    packages_to_assert = [
```
oyeb: Could you please comment which variant is going where, and the origin of these packages?
Could you also comment in each test case what were the packages _befor_ splitting? It will make the splitter's action more visible to the reader.

PR #81, opened 2018-05-23
Add shipment engine

arjun289:
## Shipment Engine
* This feature is a part of address -> payments,
* It handles the creation of packages which fulfills the given order under certain `constraints`.
* A detailed description of this feature can be found here [##3967750](https://www.pivotaltracker.com/epic/show/3967750)

What this adds to the project
- A shipment engine module to generate the graph for CSP from given packages and the discussed 
  set of constraints[##3967750](https://www.pivotaltracker.com/epic/show/3967750) and solve the CSP.
- A test suite and support for running the shipment engine tests.
- [delivers #157664486]

> Note 
The following shipment is a very basic CSP solver and doesn't covers all the constraints, we have to model our problem even better and use ArcConsistency instead of SimulatedAnnealing to come to the right solution.


apps/snitch_core/test/domain/shipment_engine_test.exs
```diff
@@ -0,0 +1,91 @@
...
+  @sl_manifest %{
+    "default" => %{
+      default: true
+    },
+    "backup" => %{},
+    "origin" => %{}
+  }
+
+  setup :variants
+  setup :stock_locations
```
oyeb: AFAICS, not required. This will be needed later if our (default) CSP uses zone info -- and when that happens we have to insert many more things as well.

mix.lock
```diff
@@ -1,6 +1,7 @@
 %{
   "abnf2": {:hex, :abnf2, "0.1.2", "6f8792b8ac3288dba5fc889c2bceae9fe78f74e1a7b36bea9726ffaa9d7bef95", [:mix], [], "hexpm"},
-  "argon2_elixir": {:hex, :argon2_elixir, "1.3.0", "fbc521ca54e8802eeaf571caf1cf385827db3b02cae30d9aa591b83ea79785c2", [:make, :mix], [{:elixir_make, "~> 0.4", [hex: :elixir_make, repo: "hexpm", optional: false]}], "hexpm"},
+  "argon2_elixir": {:hex, :argon2_elixir, "1.2.14", "0fc4bfbc1b7e459954987d3d2f3836befd72d63f3a355e3978f5005dd6e80816", [:make, :mix], [{:elixir_make, "~> 0.4", [hex: :elixir_make, repo: "hexpm", optional: false]}], "hexpm"},
```
oyeb: Why have the versions dropped down? :thinking: 
Also L29.

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -0,0 +1,152 @@
...
+  """
+
+  import Aruspex.Problem, only: [new: 0, add_variable: 3, post: 4, post: 3]
+  alias Snitch.Data.Schema.Order
+  alias Aruspex.Strategy.SimulatedAnnealing
+
+  @domain [true, false]
+
+  @doc """
+  Returns a list of packages fulfilling an order
```
oyeb:
```
@doc """
Returns the optimal shipment.

Something about what we really mean by an optimal shipment.
"""
```

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -0,0 +1,152 @@
...
+
+  Takes a list of `packages` and an `order` struct.
+  """
+  @spec run(list, Order.t()) :: list
+  def run(packages, order) do
+    edges = create_csp(packages)
+    pckgs = append_keys(packages)
```
oyeb: typo: vowels missing :see_no_evil:

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -0,0 +1,152 @@
...
+
+    vars =
```
oyeb: Why is this a `reduce`?

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -0,0 +1,152 @@
...
+    |> append_keys()
+    |> Enum.map(&group_package_with_item(&1))
+    |> List.flatten()
+    |> group_packages_by_item()
+    |> packages_with_same_item()
+    |> filter_self_edges()
+    |> find_unique_edges()
+  end
+
+  defp append_keys(packages) do
```
oyeb: You could have used [`Stream.with_index/2`](https://hexdocs.pm/elixir/Stream.html#with_index/2).

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -0,0 +1,152 @@
...
+    end)
+  end
+
+  defp item_var_mapping(packages) do
+    Enum.map(packages, fn pkg ->
+      {pkg.id, pkg}
+    end)
+  end
+
+  defp create_csp(packages) do
```
oyeb: A couple of comments:
1. `group_package_with_item/1` and `group_packages_by_item/1` sound similar but do very different things.
2. All the private functions used in this transformation are anonymous (with poorly named vars). It is better to _name those anonymous  `fn`s_ and invoke them in a `Stream` or `Enum` function.
3. This transformation can make better use of the `MapSet` in `package.variants` <- (not a change request).

apps/snitch_core/test/domain/shipment_engine_test.exs
```diff
@@ -0,0 +1,91 @@
...
+
+  describe "order fulfilled" do
+    @tag variant_count: 4
+    test "with multiple packages", context do
+      package_types = [[1, 0, 1, 0], [1, 1, 1, 1], [1, 0, 0, 0], [0, 1, 0, 1]]
```
oyeb: `package_manifest` reads better.

apps/snitch_core/test/domain/shipment_engine_test.exs
```diff
@@ -0,0 +1,91 @@
...
+      |> stock_locations_with_manifest()
+
+    {_, stock_locations} = Repo.insert_all(StockLocation, locations, returning: true)
+
+    [stock_locations: stock_locations]
+  end
+
+  defp fulfilled_items(pckgs) do
+    Enum.reduce(pckgs, [], fn %{items: list}, acc ->
+      Enum.reduce(list, acc, fn %{line_item: item}, acc -> [item | acc] end)
```
oyeb: 1. Reads much better as an `Enum.map`.
2. It is much cheaper to make 2 MapSets of variant/item ids and assert that they are equal.

apps/snitch_core/test/support/shipment_engine_case.ex
```diff
@@ -0,0 +1,69 @@
...
+      package_with_items(lineitems, sls, vs, i)
+    end
+  end
+
+  defp package_with_items(lineitems, stocks, variants, item_fulfillment) do
+    stock = Enum.random(stocks)
+
+    items =
+      lineitems
+      |> Enum.zip(variants)
```
oyeb: Use `Stream`.

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -0,0 +1,152 @@
...
+  alias Aruspex.Strategy.SimulatedAnnealing
+
+  @domain [true, false]
+
+  @doc """
+  Returns a list of packages fulfilling an order
+
+  Takes a list of `packages` and an `order` struct.
+  """
+  @spec run(list, Order.t()) :: list
```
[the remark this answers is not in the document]
arjun289: don't have `Package.t()` yet

apps/snitch_core/lib/core/domain/shipment_engine.ex
arjun289: vars is a list of id's assigned to the packages `[:p1, :p2, ...]` the

apps/snitch_core/lib/core/domain/shipment_engine.ex
arjun289: `Stream.with_index` need indexes to be an integer, here atoms are being used for identification.

apps/snitch_core/test/domain/shipment_engine_test.exs
arjun289: had thought of this but creating a `MapSet` of items from packages may lead to skipping of those cases in which lineitems may be repeated in two packages.

apps/snitch_core/lib/core/domain/shipment_engine.ex
arjun289: `group_package_with_item` => [{p1, item1}, {p2, item1}, {p3, item1}] here `package` is singular.
`group_packges_by_item` => [item1 => {p1, p2, p3}] here `packages` is plural.
They may sound similar but, mean different things.

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -1,66 +1,77 @@
 defmodule Snitch.Domain.ShipmentEngine do
   @moduledoc """
-  Implements functions to handle creation of the optimal shipment.
+  Finds the optimal shipment for a given order.
 
-  > Note 
-    No support for partially fulfilled and backordered packages.
+  The current module returns `shipments` under the following constraints:
+  * Selected `shipments` won't fulfill the same `lineitem`.
```
oyeb: We provide only _one `shipment`_. The `shipment` has many `packages`. Hence:
1. Two (or more) packages will not fulfill the same line-item
2. The shipment as a whole will fulfill the complete order.

It's more important to highlight that we don't give a damn about backorder. Please refrain from "current" module, and mention that the shipment is found using a CSP.

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -1,66 +1,77 @@
...
   @domain [true, false]
 
   @doc """
-  Returns a list of packages fulfilling an order
+  Returns the optimal shipment.
 
   Takes a list of `packages` and an `order` struct.
+  An optimal shipment which fulfills the specified business
+  constraints is returned.
+  At present `shipments` which fulfill the `order` and are mutually
```
oyeb: Confusing wording.
What specified business constraints? There's no way to state these constraints so just state what this does.

apps/snitch_core/lib/core/domain/shipment_engine.ex
oyeb: I'm just saying this could have been a `map`.

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -0,0 +1,161 @@
...
+    packages
+    |> append_keys()
+    |> Enum.map(&group_package_with_item(&1))
+    |> List.flatten()
+    |> group_packages_by_item()
+    |> packages_with_same_item()
+    |> find_unique_edges()
+  end
+
+  defp append_keys(packages) do
```
oyeb: So what if you're building atoms? isn't the "reduction" on the list of integers?
Here's how I would use `with_index`:
```elixir
packages
|> Stream.with_index()
|> Enum.map(fn {package, index} ->
  Map.put(package, :id, String.to_atom("p#{index}"))
end)
```

mix.lock
oyeb: Please fix your `mix.lock`. You might have to do `mix deps.update --all`

apps/snitch_core/test/domain/shipment_engine_test.exs
arjun289: but that is a guarantee given by this module and, shouldn't that be tested?

oyeb: No, this guarantee is given by `Snitch.Domain.Shipment`, the module that _creates_ the list that is _simply being consumed here,_ in `ShipmentEngine`. 
Hence, you can happily use the `MapSet`.

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -0,0 +1,146 @@
...
+  Uses [Simulated Annealing][sa] to find the optimal shipment, see
+  `Aruspex.Strategy.SimulatedAnnealing`. Note that this technique is
+  probabilistic and [AC-3][ac3] should replace it.
+
+  [sa]: https://en.wikipedia.org/wiki/Simulated_annealing
+  [ac3]: https://en.wikipedia.org/wiki/AC-3_algorithm
+  """
+  @spec run(list, Order.t()) :: list
+  def run(packages, %Order{} = order) do
+    packages_with_id = append_keys(packages)
```
SagarKarwande: `append_keys` is called in `create_csp`.
Is there a need to call it again?

apps/snitch_core/lib/core/domain/shipment_engine.ex
```diff
@@ -0,0 +1,146 @@
...
+  defp variable_assignment([]), do: []
+
+  defp variable_assignment([result]) do
+    Enum.reject(result.binding, fn
+      {{:hidden, _, _}, _} -> true
+      _ -> false
+    end)
+  end
+
+  defp filter_packages(pckgs, bindings) do
```
SagarKarwande: Typo `pckgs`

apps/snitch_core/lib/core/domain/shipment_engine.ex
arjun289: I need the packages with id to create a list of vars for the CSP.

PR #88, opened 2018-05-30
Create Order API

arjun289:
* adds an API to create a new order.
   - allows the creation of order with the
     provided params.
   - if `line_items` not provided in params
    creates an empty order.
* [delivers #157749620]


apps/api/lib/api_web/controllers/order_controller.ex
```diff
@@ -0,0 +1,44 @@
+defmodule ApiWeb.OrderController do
+  use ApiWeb, :controller
+
+  alias Snitch.Data.Model.Order
```
SagarKarwande: Group the aliases `Snitch.Data.Model.{Order, User}`

PR #92, opened 2018-05-31
Add Checkouts controller with cart-to-address

oyeb:
## Motivation
Pivotal story: [#157750610](https://www.pivotaltracker.com/story/show/157750610)

### Checkouts controller
1. PUT `checkouts/:order_id/next.json`
  spree/angular needs it for some reason
  as of now it just returns an empty 200
2. PUT `checkouts/:order_id.json`
  accepts the billing and shipping address params
  triggers the `add_addresses` event in Order state machine
  returns all the packages and their items in a spree compatible format.

### Misc
Added a view for HTTP 400.


oyeb:
Thanks for testing this everyone!

PR #94, opened 2018-06-01
Add address-to-payment in checkouts controller

oyeb:
## Motivation
Pivotal story: [#157751042](https://www.pivotaltracker.com/story/show/157751042)
Allows an order to be paid for via the "check" payment method.

## Description
* New `CheckPayment` model. Makes it simpler to create a check payment.
* Implements the `add_payment` event transition in the `Order.DefaultMachine`
  - supports `CheckPayment` exclusively!

### DIFF
--- a/apps/api/lib/api_web/controllers/product_controller.ex
+++ b/apps/api/lib/api_web/controllers/product_controller.ex
@@ -14,7 +14,7 @@ defmodule ApiWeb.ProductController do
 
   def show(conn, params) do
     product =
-      Repo.get_by(Product, slug: params["product_slug"])
+      Repo.get_by(Product, slug: params["id"])
       |> Repo.preload(variants: [:images])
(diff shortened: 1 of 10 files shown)


SagarKarwande:
⊂( ◜◒◝ )⊃

apps/api/lib/api_web/controllers/product_controller.ex
```diff
@@ -12,7 +12,7 @@ defmodule ApiWeb.ProductController do
     render(conn, "products.json", products: products)
   end
 
-  def product(conn, params) do
+  def show(conn, params) do
     product =
       Repo.get_by(Product, slug: params["product_slug"])
```
SagarKarwande: As the router has been changed to resources product slug will be in `params["product_slug"]`

PR #105, opened 2018-06-20
Seed Zone, ShippingMethod, ShippingCategory, etc

oyeb:
## Motivation
Pivotal story: [#157925979](https://www.pivotaltracker.com/story/show/157925979)

The painstakingly created `Snitch.*Case` test helpers were not available in `dev`/`prod` environments, this PR moves all of them into `helpers/` to make seeding seem like a refreshing walk in the park. :wind_chime: :hibiscus: 

## Description
* Adds various seed data: `Zone`, `ShippingMethod`, `ShippingCategory`, `StockItems` and `Locations`.
* The methods have been made fail-safe and smart, so you can run the seed task multiple times without worrying about duplicating any records.
* The slowest seed task is taxonomy.
* Fixed `Shipment.to_package` and `:number` generation in `Package`, `PackageItem` schema.
* Also fixes a not so common bug in ShipmentEngine.
  - For it to work properly, the order's `line_items` must be loaded.

### Migrations
* StockLocation `:admin_name`s must be unique.
* `Orders` can be deleted via `Repo.delete`. Earlier, an FK constraint with the `Address` would cause issues.

------

## Checklist

------
Dear Gatekeeper,

branch are [formatted according to our standards](commit-template).

> The only additional requirement is that the PR number must appear in the

[commit-template]: https://github.com/aviabird/snitch/blob/develop/.git_commit_msg.txt


apps/snitch_core/priv/repo/seed/shipping.ex
```diff
@@ -0,0 +1,72 @@
...
+      "regular" => {[india, north_india], all_categories},
+      "hyperloop" => {[north_india], [light, fragile]}
+    }
+
+    Enum.map(manifest, fn {name, {zones, categories}} ->
+      ShippingMethod.create(%{name: name, slug: name}, zones, categories)
+    end)
+  end
+
+  def seed_shipping_categories! do
```
arjun289: This function shows it will raise in case of bad query, but I don't think the query here is going to be bad, what's your thoughts, at the max this will give [no_return].

oyeb: Come on let me indulge a bit, and let the name have a `!` in the end. Can you guarantee that the call to `shipping_categories_with_manifest/1` will never raise?
The `!` indicates that the output is not guaranteed to be "graceful" -- we don't return `{:ok, term} | {:error, term}` tuples here.

PR #111, opened 2018-06-25
Transition address delivery 158384345

kanishkablack:
Transition of state from `address` to `delivery`

- Associate Package id with ShippingMethod id.
- Persists package data along with updated total.
- ~~Update Order total with package total~~
- ~~Persist Order.~~

## Issues

- ~~Blocked by `Money.zero` in test case.~~
- ~~`Mox` not able to fetch default currency.~~

------

## Checklist

------
Dear Gatekeeper,


apps/snitch_core/lib/core/data/schema/order.ex
```diff
@@ -93,7 +96,14 @@ defmodule Snitch.Data.Schema.Order do
       |> get_field(:line_items)
       |> LineItemModel.compute_total()
 
-    total = Enum.reduce([item_total], &Money.add!/2)
+    packages_total =
+      if order_changeset.data.id != nil do
```
kanishkablack: Reason to check `order_id` to `nil`, there are few test cases, where order is generated with out `id`.
I am not sure how its possible

apps/snitch_core/lib/core/data/schema/order.ex
```diff
@@ -58,6 +60,7 @@ defmodule Snitch.Data.Schema.Order do
     |> unique_constraint(:number)
     |> foreign_key_constraint(:user_id)
     |> cast_assoc(:line_items, with: &LineItem.create_changeset/2, required: true)
+    |> cast_assoc(:packages)
```
oyeb: No. We never want to changes packages via order.

apps/snitch_core/lib/core/data/schema/order.ex
oyeb: We don't want to manage packages from an order.

apps/snitch_core/lib/core/domain/order/default_machine.ex
```diff
@@ -57,16 +57,16 @@ defmodule Snitch.Domain.Order.DefaultMachine do
...
     event(:add_addresses, %{from: [:cart], to: :address}, fn context ->
       context
       |> Transitions.associate_address()
       |> Transitions.compute_shipments()
       |> Transitions.persist_shipment()
     end)
 
-    event(:add_payment, %{from: [:address], to: :payment}, fn context ->
-      context
+    event(:process_packages, %{from: [:address], to: :delivery}, fn context ->
```
oyeb: `add_shipping_preference` is a better name for the event.

apps/snitch_core/lib/core/domain/order/transitions.ex
```diff
@@ -124,4 +125,38 @@ defmodule Snitch.Domain.Order.Transitions do
   end
 
   def persist_shipment(%Context{valid?: false} = context), do: context
+
+  defp extract_shipping_method_cost(shipping_method_id) do
+    shipping_method = Repo.get(ShippingMethod, shipping_method_id)
```
oyeb: Why involve `Repo`? The package struct already has `shipping_methods`, a list of `shipping_method` embeds, each of them has the cost in it.
I would do this:
```elixir
defp extract_shipping_method_cost(package) do
  sm = Enum.find(package.shipping_methods, fn %{shipping_method_id: id} ->
    id == package.shipping_method_id
  end)
  sm.cost
end

File: apps/snitch_core/lib/core/domain/order/transitions.ex [hunk 5]
```diff
@@ -124,4 +125,38 @@ defmodule Snitch.Domain.Order.Transitions do
...
+
+  @doc """
+  Persists shipping_method_id to packages
+
+  Calculate pacakge total cost Sum(shipping_cost, adjustment_total,promo_total, shipping_cost)
+
+  Update package cost total in DB
+  """
+
+  @spec associate_package(Context.t()) :: Context.t()
```
oyeb: 1. The packages are already associated (with the order). Come up with a better name.
2. The function spec does not match what the function actually returns, it returns a list of contexts. Why?
3. The `Enum.map` is dangerous as errors are not handled. Please take a look how `persist_shipment` safely traverses a list of repo operations, in a `Multi`.

apps/snitch_core/lib/core/data/model/package.ex
```diff
@@ -36,4 +37,37 @@ defmodule Snitch.Data.Model.Package do
...
+  def compute_package_total(order) do
+    case get_packages(order) do
+      [] ->
+        MoneyTools.zero!()
+
+      packages ->
+        packages
+        |> Stream.map(&Map.fetch!(&1, :total))
+        |> Enum.reduce(&Money.add!/2)
+        |> Money.reduce()
```
oyeb: Unnecessary

apps/snitch_core/lib/core/data/schema/order.ex
oyeb: The order `:id` is set only when the struct is inserted into the DB. Many test cases don't touch the repo.

apps/snitch_core/lib/core/domain/order/default_machine.ex
kanishkablack: 👍

apps/snitch_core/test/domain/order/transitions_test.exs
```diff
@@ -141,4 +142,20 @@ defmodule Snitch.Domain.Order.TransitionsTest do
...
     end
   end
+
+  describe "associate_package" do
+    setup do
+      [order: insert(:order, user: build(:user))]
+    end
+
+    test "when packages are empty", %{order: order} do
+      result =
```
oyeb: You will have to call `Mox.expect` here. Line 155 will eventually result in a call to `MoneyTools.zero`, which calls the `Tools.Defaults.fetch`, to fetch the default currency.
Refer the `Snitch.Tools.MoneyTest` module.

apps/snitch_core/lib/core/data/schema/order.ex
kanishkablack: Packages are not updated or edited from Order,
To update total of order, packages are fetched to get the packages costs.

apps/snitch_core/lib/core/domain/order/transitions.ex
kanishkablack: Context data should be struct `preloaded` before sending to state machine, if I am not wrong ?

apps/snitch_core/lib/core/data/schema/order.ex
kanishkablack: To involve `package module` in order to calculate `total` of order.

I didn't see the story - https://www.pivotaltracker.com/n/projects/2149807/stories/158605364

oyeb: If you don't want to edit packages from an order, then why is there a `cast_assoc(:packages)`?

apps/snitch_core/lib/core/domain/order/transitions.ex
oyeb: You are once again thinking about _triggering the state machine_. Please work on _handling the trigger_ instead.
Moreover, I don't see why you are fetching shipping methods from the DB, **when you already have them, and the costs as well, in the `package`**.

apps/snitch_core/lib/core/domain/order/transitions.ex
```diff
@@ -124,4 +124,54 @@ defmodule Snitch.Domain.Order.Transitions do
...
+  def associate_package(%Context{valid?: true, struct: %Order{} = order} = context) do
```
oyeb: Where are the `form_params` that we discussed about (so clearly) in the [meeting](https://hackmd.io/atLBxkpvReiEwSbx5CzUTA#Live-coding)? Shouldn't they be a part of the context?
Why is this function named `associate_packages`? The packages are already associated with the order :thinking:

apps/snitch_core/lib/core/domain/order/transitions.ex
```diff
@@ -124,4 +124,54 @@ defmodule Snitch.Domain.Order.Transitions do
...
   def persist_shipment(%Context{valid?: false} = context), do: context
+
+  defp calculate_package_total(
+         %{tax_total: tax_total, adjustment_total: adjustment_total, promo_total: promo_total},
+         shipping_cost
+       ) do
+    Enum.reduce([tax_total, adjustment_total, promo_total, shipping_cost], &Money.add!/2)
+  end
+
+  def process_package(package) do
```
oyeb: IMO this should be a private function.
This function assumes that the `package.shipping_method_id` is already set, see L138.
Then the same `shipping_method.id` is set as the `shipping_method_id` in L148. :confused: 

Since this PR moves us to `delivery` state (clearly defined in the [meeting](https://hackmd.io/atLBxkpvReiEwSbx5CzUTA#Live-coding) and [here](https://hackmd.io/4Neif6WUQIGSftYPpiMMTw?view#States-of-an-Order)), we should be "setting" the various package total fields (and not just reading them).
You are summing the package totals, but we know for a fact that `tax_total`, `adjustment_total` and `promo_total` are `nil`s.

apps/snitch_core/test/domain/order/transitions_test.exs
```diff
@@ -141,4 +148,36 @@ defmodule Snitch.Domain.Order.TransitionsTest do
...
+          order_id: order.id,
+          origin: build(:stock_location),
+          shipping_methods: build_list(1, :embedded_shipping_method),
+          shipping_category: build(:shipping_category)
+        )
+
+      [order: order, packages: packages]
+    end
+
+    test "with packages", %{order: order, packages: packages} do
```
oyeb: The setup built here does not accurately mimic the real scenario. It was made crystal clear in the [meeting](https://hackmd.io/atLBxkpvReiEwSbx5CzUTA#Live-coding) that when the transition is called, the `package.shipping_method_id` is `nil`. You can't set the `:id` like this on L170.
Moreover, what is the significance of this `3`?

apps/snitch_core/lib/core/data/schema/order.ex
```diff
@@ -55,7 +56,7 @@ defmodule Snitch.Data.Schema.Order do
     order
     |> cast(params, @create_fields)
     |> validate_required(@required_fields)
-    |> unique_constraint(:number)
+    |> unique_constraint(:slug)
```
oyeb: This should be `:number`, might have missed it during some merge resolution.

apps/snitch_core/lib/core/domain/order/default_machine.ex
```diff
@@ -57,16 +57,17 @@ defmodule Snitch.Domain.Order.DefaultMachine do
...
+                ~w(cart address payment delivery processing rts shipping complete cancelled)a do
     event(:add_addresses, %{from: [:cart], to: :address}, fn context ->
       context
       |> Transitions.associate_address()
       |> Transitions.compute_shipments()
       |> Transitions.persist_shipment()
     end)
 
-    event(:add_payment, %{from: [:address], to: :payment}, fn context ->
+    event(:save_packages_methods, %{from: [:address], to: :delivery}, fn context ->
```
oyeb: IMO `add_shipping_preferences` was fine, `save_shipping_preferences` is also fine.

apps/snitch_core/lib/core/domain/order/transitions.ex
```diff
@@ -124,4 +127,57 @@ defmodule Snitch.Domain.Order.Transitions do
...
 
   def persist_shipment(%Context{valid?: false} = context), do: context
+
+  defp process_package(package, shipping_method_id) do
+    shipping_method =
+      Enum.find(package.shipping_methods, fn %{id: id} ->
+        id == shipping_method_id
+      end)
+
+    package_total =
```
oyeb: Alongwith the `package_total`, we must compute the `tax_total`, `promo_total` and `adjustment_total` as well. Just set them to zeroes.
When the relevant `Domain` functions are added, we'll call them here.

apps/snitch_core/lib/core/domain/order/transitions.ex
```diff
@@ -124,4 +127,57 @@ defmodule Snitch.Domain.Order.Transitions do
...
+        %Context{valid?: true, struct: %Order{id: order_id} = order} = context
+      ) do
+    %{state: %{selected_shipping_methods: selected_shipping_methods}, multi: multi} = context
+
+    packages = Repo.all(from(p in PackageSchema, where: p.order_id == ^order_id))
+
+    function = fn _ ->
+      selected_shipping_methods
+      |> Stream.map(fn %{package_id: package_id, shipping_method_id: shipping_method_id} ->
+        package = Enum.find(packages, fn package -> package.id == package_id end)
```
oyeb: I wouldn't mind doing `package = Package.get(package_id)` here (and not fetch them together), but it's fine.
More importantly, I'd either,
* like you to do some validations here (or somewhere):
  1. ensure that a `shipping_method` is chosen for all `packages` of this `order`.
  2. ensure that `package_ids` are unique (in the `selected_shipping_methods` map).
  3. what else is needed here @pkrawat1, @ashish173, @arjun289, @SagarKarwande ?
* like you to document assumptions about the "correctness" of this `map`, so that the validation can be written later/elsewhere.

apps/snitch_core/lib/core/domain/order/transitions.ex
```diff
@@ -124,4 +127,57 @@ defmodule Snitch.Domain.Order.Transitions do
   end
 
   def persist_shipment(%Context{valid?: false} = context), do: context
+
+  defp process_package(package, shipping_method_id) do
+    shipping_method =
```
oyeb: The result of `Enum.find` could be `nil`. We must raise an error in that case.
Overall, there is a need to validate the form-submission. @ashish173 can we assume that the validation is done by the controller?
I'm skeptical of that because that _may_ couple the controller with `snitch_core`... You guys decide.

apps/snitch_core/test/support/factory/shipping.ex
```diff
@@ -13,6 +14,16 @@ defmodule Snitch.Factory.Shipping do
         }
       end
 
+      def embedded_shipping_method_factory do
+        %EmbeddedShippingMethod{
+          id: sequence(:id, &(&1 + 1)),
+          slug: sequence("shipping_method"),
+          name: sequence(:name, &"method-#{&1}"),
+          description: "Snitch revolution",
```
oyeb: :blush:

apps/snitch_core/test/support/factory/shipping.ex
```diff
@@ -34,7 +45,8 @@ defmodule Snitch.Factory.Shipping do
           total: Money.new(0, :USD),
           tax_total: Money.new(0, :USD),
           adjustment_total: Money.new(0, :USD),
-          promo_total: Money.new(0, :USD)
+          promo_total: Money.new(0, :USD),
+          id: sequence(:id, &(&1 + 1))
```
oyeb: This is auto-generated (upon insert).

apps/snitch_core/lib/core/domain/order/transitions.ex
ashish173: we agreed on putting this inside the controllers for now and have a wrapper with 2 functionalities `common api` & `validations`, TBD later once we have more transitions.

apps/snitch_core/lib/core/domain/order/default_machine.ex
ashish173: yes, the naming should be around shipping methods. `shipping` word in the name is really important here.

apps/snitch_core/lib/core/domain/order/transitions.ex
```diff
@@ -124,4 +127,57 @@ defmodule Snitch.Domain.Order.Transitions do
...
+  Update package cost total in DB
+  """
+
+  @spec save_packages_methods(Context.t()) :: Context.t()
+  def save_packages_methods(
+        %Context{valid?: true, struct: %Order{id: order_id} = order} = context
+      ) do
+    %{state: %{selected_shipping_methods: selected_shipping_methods}, multi: multi} = context
+
+    packages = Repo.all(from(p in PackageSchema, where: p.order_id == ^order_id))
```
ashish173: @kanishkablack can we have a method for this like `order_packages` and extract this from here to `orders` model?

apps/snitch_core/lib/core/domain/order/transitions.ex
ashish173: @oyeb most important is the first validation that you pointed out. we are good with these for now.

apps/snitch_core/lib/core/domain/order/transitions.ex
kanishkablack: Sure

apps/snitch_core/lib/core/domain/order/default_machine.ex
kanishkablack: `save_shipping_preferences` is appropriate

apps/snitch_core/lib/core/data/model/package.ex
```diff
@@ -36,4 +37,35 @@ defmodule Snitch.Data.Model.Package do
...
+        |> Stream.map(&Map.fetch!(&1, :total))
+        |> Enum.reduce(&Money.add!/2)
+    end
+  end
+
+  @doc """
+  Query of packages related to order
+  """
+  @spec get_packages(Order.t()) :: list(Package.t())
+  def get_packages(order) do
```
pkrawat1: how about `packages_for_order()` as fn name. 

Also can we use use OrderModel.order_packages(order) in this function, to avoid repeatitive changes that might come in future or having only one function either in package or order.

apps/snitch_core/lib/core/data/schema/order_address.ex
```diff
@@ -0,0 +1,123 @@
...
+    |> assoc_state_and_country()
+  end
+
+  defp struct_to_map(struct) do
+    struct
+    |> Map.from_struct()
+    |> Map.delete(:__meta__)
+  end
+
+  defp assoc_state_and_country(
```
pkrawat1: @oyeb we have similar code in multiple places for address ? Will have to see if it can be extracted to an address validator across the system.

apps/snitch_core/lib/core/domain/order/transitions.ex
```diff
@@ -124,4 +128,60 @@ defmodule Snitch.Domain.Order.Transitions do
...
+      end)
+
+    package_total =
+      Enum.reduce(
+        [shipping_method.cost],
+        &Money.add!/2
+      )
+
+    Package.update(package, %{
+      cost: shipping_method.cost,
```
pkrawat1: This would give error on `shipping_method.cost` if `shipping_method` is nil. This can't be handled in controller. It would require `try rescue block` then.

apps/snitch_core/lib/core/domain/order/transitions.ex
```diff
@@ -124,4 +128,60 @@ defmodule Snitch.Domain.Order.Transitions do
...
+      total: package_total,
+      tax_total: MoneyTools.zero!(),
```
pkrawat1: can we put `MoneyTools.zero!()` value in a variable.

apps/snitch_core/lib/core/data/model/package.ex
oyeb: You're right, this function might not be needed:
```elixir
packages = Map.fetch!(Repo.preload(order, [:packages]), :packages)
```
Best way to get packages of an order.

apps/snitch_core/lib/core/data/schema/order_address.ex
oyeb: I think we can get rid of the existing table-backed Address schema altogether, so we'll have only this module in the end.

pkrawat1: `existing table-backed Address` ? 

We'll need the same functions in stock locations also. and user address too. So we can extract state country validation to a validator.

oyeb: No, we don't. We'll just embed `Snitch.Data.Schema.OrderAddress` everywhere (after renaming it to `EmbeddedAddress` perhaps).

pkrawat1: hmm

apps/snitch_core/lib/core/domain/order/transitions.ex
kanishkablack: Yes, we need guard functions in context or wrapper in web controller.
TMB, we are moving ahead.

apps/snitch_core/test/domain/order/transitions_test.exs
```diff
@@ -141,4 +149,49 @@ defmodule Snitch.Domain.Order.TransitionsTest do
...
+      ]
+
+      result =
+        order
+        |> Context.new(state: %{selected_shipping_methods: shipping_methods})
+        |> Transitions.save_packages_methods()
+
+      assert result.valid?
+      assert [packages: {:run, _}] = Multi.to_list(result.multi)
+      assert {:ok, %{packages: _}} = Repo.transaction(result.multi)
```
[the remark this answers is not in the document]
kanishkablack: 1. Added

2. I think we can never arrive at this scenario, order without packages, at delivery.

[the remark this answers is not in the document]
oyeb: Unfortunately, (2) is possible. Say an `order` is in `address` state and the user goes to the cart and empties it. According to [the epic I'm working on], we'll still keep the order is `address` state, and this order has no packages.
There are some other scenarios where this can happen as well.

PR #127, opened 2018-07-05
 Remove the line_item.total field

oyeb:
## Motivation
We must remove fields that simply "cache" computations from the DB. They
rarely help (optimization-wise) but become a huge PITA when we update
the "operands" of said computation -- the update must cascade to these
cache fields as well.

## Description
This change gets rid of `line_item.total` field.

* Compute the total inline in all places that required this field.
* Added tests for `Model.Variant`.
* Updated the views in `snitch_api` and `admin_app`

Migrations
----------
Remove `:total` from `LineItem`.

------

## Checklist

------
Dear Gatekeeper,

### DIFF
--- a/apps/snitch_core/lib/core/data/model/variant.ex
+++ b/apps/snitch_core/lib/core/data/model/variant.ex
@@ -3,10 +3,22 @@ defmodule Snitch.Data.Model.Variant do
   Variant API
   """
   use Snitch.Data.Model
+
   alias Snitch.Data.Schema.Variant
 
   def get_category(%Variant{} = v) do
     variant = Repo.preload(v, :shipping_category)
     variant.shipping_category
+  end
+
+  def get_selling_prices(variant_ids) do
+    # TODO: change the source of selling price to the something else.
+    query = from(v in Variant, select: {v.id, v.selling_price}, where: v.id in ^variant_ids)
+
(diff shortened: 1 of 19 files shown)


apps/snitch_core/lib/core/data/model/variant.ex
```diff
@@ -3,10 +3,22 @@ defmodule Snitch.Data.Model.Variant do
   Variant API
   """
   use Snitch.Data.Model
+
   alias Snitch.Data.Schema.Variant
 
   def get_category(%Variant{} = v) do
```
ashish173: @oyeb IMO this method should be renamed as `get_shipping_category`. It is not obvious from the name of the function which category it returns.

apps/snitch_core/test/data/model/variant_test.exs
```diff
@@ -0,0 +1,34 @@
+defmodule Snitch.Data.Model.VariantTest do
+  use ExUnit.Case, async: true
+  use Snitch.DataCase
+
+  import Snitch.Factory
+
+  alias Snitch.Data.Model.Variant
+
+  test "get_category/1" do
```
ashish173: @oyeb this test should become changes as per my above change suggestion in the name of the function.

PR #129, opened 2018-07-06
Products pagination-search-filter-sort 158821938

blackode:
**Product search/filter using seed data**
## Motivation and Context
The products pagination, search, and filters are agenda for this PR. We list out the products according to the query parameters. Provided an option for filtering data based on the `name` sub-text.

Filtering, at present `%{"filter" => %{"name" => name}}` is only implemented. However, this is scalable. 
the **name** is not fully matched to extend it to further. 

Pagination links like `self` `next`, `prev`, `last` are developed according to JSON-API spec. However,
not provided with page parameters, can override the page params.

[Pivotal Story Link](https://www.pivotaltracker.com/story/show/158821938)

## Things covered in this PR

- API followed JSON-API spec
- API with pagination support
- Pagination with default page setup
- Pagination with overriding page feature   
- Product filter search `A-z|Z-A|Newest{inserted_at}`
- Products filter by name contains the search `substring(ilike)`
## Further comments
For the sake of convenience, I took the page size as 2. We can update it in `config.exs` under `ja_serializer` app of key `page_size`. This key was added manually. It is not present in the serializer.

The implemented feature unit test cases are written with some unusual insertions of rows into DB. :wow:

------

## Checklist

------
Dear Gatekeeper,

> The only additional requirement is that the PR number must appear in the


apps/snitch_api/lib/snitch_api_web/controllers/product_controller.ex
```diff
@@ -0,0 +1,106 @@
...
+  use SnitchApiWeb, :controller
+
+  alias Snitch.Repo
+  alias Snitch.Data.Schema.Product
+
+  import Ecto.Query, only: [from: 2]
+
+  # sort the products in Z-A descending order_by prodcut name
+  def index(conn, %{"sort" => "Z-A"} = params) do
+    query = from(p in Product, order_by: [desc: p.name])
```
pkrawat1: Why no use of domain functions. Why import Ecto here ?

blackode: For developing non destructive queries

pkrawat1: @blackode How about we create some context to wrap these functions can call them from here.

pkrawat1: @blackode use context, fallback controller that phoenix 1.3 + provides.

apps/snitch_api/lib/snitch_api_web/views/product_view.ex
```diff
@@ -0,0 +1,33 @@
...
+  ])
+
+  has_many(
+    :variants,
+    serializer: SnitchApiWeb.VariantView,
+    include: true
+  )
+
+  def variants(product, _conn) do
+    Map.get(product, :variants)
```
pkrawat1: Why was this required. Was it not coming from the schema. I might be missing something. Can you tell.

blackode: That is just for safety. By default, it implements the function of every attribute. For an example, In our `Image` schema we have an attribute  called `url`. The `ja_serializer` defines the function `url/2` where there is alredy one defined from `conn` which gives conflicts. So, I renamed the attribute with `image_url` instead of `url` and overriding the `def image_url/2` function which extracts `url`.

apps/snitch_api/lib/snitch_api_web/controllers/product_controller.ex
```diff
@@ -0,0 +1,106 @@
...
+
+  defp gen_page_links(collection, params, conn) do
+    {number, size} = extract_page_params(params)
+
+    JaSerializer.Builder.PaginationLinks.build(
+      %{
+        number: number,
+        size: size,
+        total: div(Enum.count(collection), size),
+        base_url: "http://0.0.0.0:3000/api/v1/products"
```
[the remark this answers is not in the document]
blackode: It returns the only `host`

apps/snitch_api/lib/snitch_api_web/controllers/product_controller.ex
```diff
@@ -0,0 +1,78 @@
+defmodule SnitchApiWeb.ProductController do
+  use SnitchApiWeb, :controller
+
+  alias SnitchApi.API, as: Context
+
+  action_fallback(SnitchApiWeb.FallbackController)
+
+  # sort the products in Z-A descending order_by prodcut name
+  def index(conn, %{"sort" => "Z-A"} = params) do
+    {products, page} = Context.products_by_desc(params, conn)
```
pkrawat1: Have you tested error case for the actions, like are you getting fallback responses on error?

PR #180, opened 2018-09-10
Order confirmation mail

jyotigautam:
**Why?**
The user should be able to receive an order confirmation mail after its completion.

**This change addresses the need by:**
Feature:

*  An email sent to the user showing the particulars of the order made by him after completing the order.


apps/admin_app/lib/admin_app_web/controllers/session_controller.ex
```diff
@@ -143,13 +143,14 @@ defmodule AdminAppWeb.SessionController do
   end
 
   defp verify_password_token(token) do
-    max_age = System.get_env("token_maximum_age") |> String.to_integer()
+    max_age = System.get_env("token_maximum_age")
```
arjun289: Instead of getting this directly from the `system env` we can put this in the `config` and there you can pull it from the system.

PR #245, opened 2018-10-08
Multitenancy

ramansah:
## Requirements
1. Multiple tenants (subdomains)
Example : amazon.aviacommerce.com & flipkart.aviacommerce.com running on same machine with DB level data separation
Terminology : Amazon and Flipkart are “tenants”. (like tenants in an apartment // server)
2. Tables will still have root level data. E.g. website will still host products on root level domain, aviacommerce.com
3. Utility for adding / removing tenants - UI based. (Do we need this)

## Doubts
* Level of customization for each tenant.
E.g. Logo & website name different for each tenant?
(If no customization, then “Implementation -2” will be avoided)
* Domain structure.
If enabled, will both frontend and backend coexist in same domain?
Then hosts would look like api.amazon.aviacommerce.com

## Implementation : Use Triplex library
https://github.com/ateliware/triplex
Reason : Ready to use methods for adding / removing tenants, supports PostgreSQL & MySQL. Least intrusive to Ecto, snitch will have its own implementation for data manipulation with Ecto.

Normal operations at root level domain will continue to work and multi tenancy will be active by default. New subdomain will become accessible only if admin creates a tenant.

- [x] Common plug used by admin_app and snitch_api to parse subdomain
- [x] Repo operations : Repo.insert(user, prefix: “amazon”)
If prefix: nil, data inserted to general table
- [x] Some migrations need schema prefix
create function #{ prefix() || "public" }.zone_exclusivity
- [x] Changes in Triplex
Modify “tenant_migrations_path” to point to default migrations path so that all tables are multi tenant and all migrations go under it. No need to move our migrations to “tenant_migrations” directory.
- [x] 404 if subdomain / tenant not present

### Better Design / Improvements Needed
1. Passing tenant in subdomain from plug to repo
Majority of code change would be converting Repo.get(:id, User) to Repo.get(:id, User, prefix: “amazon”) and Users.get_all() to Users.get_all(schema: “amazon”)
-  [x] Solution part 1
Each request has its own process, so plug will store subdomain in process dict and repo functions will retrieve it later, eliminating need to rewrite models and controllers.
https://elixirforum.com/t/arch-question-tired-of-passing-this-same-parameter-around-for-every-method-call/2166/4
- [x] Solution part 2 
Create Snitch.MultiTenant.Repo which will implement DB CRUD operations by retrieving prefix from process dict. This will ensure no changes in Repo methods’ signature. Only need to change aliases in controllers, models, tests. 
Config file would still reference Snitch.Repo along with supervisor.

2. Caching strategy for tenant info : W.I.P, depends on Doubt -1
For every request, an additional DB call to fetch tenant info from slug (loading logo, name)

## Checklist
- [ ] Sync repository for new commits

## How To Test
* Edit hosts file to map localhost to a dummy domain
`sudo nano /etc/hosts`
```
127.0.0.1       tenant1.mysite.com
```
* Create a tenant in iex console `Triplex.create("tenant1")`
* Create an admin user `Repo.insert(user, prefix: "tenant1")`
* Browse to `tenant1.mysite.com` to access the app

Or run all tests against a tenant

```
cd apps/snich_core
mix test.multi
```

------
Dear Gatekeeper,


apps/snitch_core/priv/repo/migrations/20180213103123_add_money_with_currency_type_to_postgres.exs
```diff
@@ -3,11 +3,11 @@ defmodule Core.Repo.Migrations.AddMoneyWithCurrencyTypeToPostgres do
 
   def up do
     execute """
-      CREATE TYPE public.money_with_currency AS (currency_code char(3), amount numeric(20,8))
+      CREATE TYPE #{ prefix() || "public" }.money_with_currency AS (currency_code char(3), amount numeric(20,8))
```
pkrawat1: @ramansah How can I run this on systems already with previous migration in DB.
We never update a migration, instead we create another migration to update the DB.

PR #278, opened 2018-10-17
Shipping rules

arjun289:
Manage Shipping policy

### Story Link
https://www.pivotaltracker.com/story/show/161209703
The shipping cost for an order is calculated based on some rules. These rules are set and managed
by shipping policy in the system.

## Motivation and Context

The admin of a store should be able to set up some rules for shipping category in the store. When 
an order is placed and the packages are created for the order the shipping cost of the package can be calculated based on the shipping rules.
The shipping policy is modeled in the following manner
```
ShippingCategory  ---> has_many ---> ShippingRules
```
Out of all the rules that can be activated for a shipping_category, only
    one out of group relating to some entity can be selected.
    Let's say the shipping rules are:
    **product**
      - flat rate for each product
      - different rate for each product
     **order**
      - fixed rate for all orders
      - free above some amount for an order

 Out of the above, one rule can be picked from `product` and, one
    from `order`.

## Describe your changes

- have added schema and migrations for the new entities `shipping identifier` and `shipping rule`
- also added a module which would perform the calculations
- frontend for managing the policies

## Any further comments?
This is a POC and there are going to be modification and further rule additions in further PRs.
 

------

## Checklist

------
Dear Gatekeeper,


apps/admin_app/lib/admin_app_web/router.ex
```diff
@@ -94,6 +94,11 @@ defmodule AdminAppWeb.Router do
       ProductController,
       :delete_property
     )
+
+    # get("/shipping-policy", ShippingPolicyController, :index)
```
ashish173: @arjun289 can we remove the commented code?

arjun289: will remove this

apps/admin_app/assets/js/views/shipping_policy/edit.js
```diff
@@ -0,0 +1,69 @@
+import MainView from '../main';
+
+export default class View extends MainView {
```
arjun289: @anantanant2015 can you please make this big piece of a code a function and call it in the two files. This will remove code repetition.

PR #328, opened 2018-11-26
Display active product with state.

anantanant2015:
Display product based on state field.

## Motivation and Context

- Product displayed using field is_active, and should be displayed by using
  state of the product.
- Deletion of product on success should update state as "deleted", not by the 
  field is_active set to false.
- Remove field based filter methods from product context for taxon, brand
  and slug should use list products method to filter product.

## Describe your changes

- Updates query for state, and remove is_active field from
  query.
- Removes is_active field in product schema delete changeset
  and use state field.
- Removes independent methods to get product by taxon,
  brand.
- Removes index action for independent methods for brand, taxon.
- Updates show action in product controller to use list products
  method.
--------

## Checklist

------
Dear Gatekeeper,

### DIFF
--- a/apps/snitch_api/lib/snitch_api_web/controllers/product_controller.ex
+++ b/apps/snitch_api/lib/snitch_api_web/controllers/product_controller.ex
@@ -48,17 +48,17 @@ defmodule SnitchApiWeb.ProductController do
   end
 
   def show(conn, %{"product_slug" => slug} = params) do
-    case Context.list_products(conn, params |> Map.put("filter", %{"slug" => slug})) do
-      {[], _page} ->
+    case Context.product_by_slug(slug) do
+      {:error, :not_found} ->
         conn
         |> put_status(:not_found)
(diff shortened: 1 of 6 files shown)


apps/snitch_api/lib/snitch_api_web/controllers/product_controller.ex
```diff
@@ -71,17 +48,17 @@ defmodule SnitchApiWeb.ProductController do
   end
 
   def show(conn, %{"product_slug" => slug} = params) do
-    case Context.product_by_slug(slug) do
```
ashish173: @anantanant2015 this change is not required.

ashish173: `show` method should be just used to fetch a single product. We should not confuse this with the filtering api and not try to make it common with context methods. Please make these changes.

anantanant2015: done

PR #329, opened 2018-11-28
 list for currency selection in general settings

jyotigautam:
Adding a dropdown on create general settings form to choose a currency from a given list of currencies.

## Motivation and Context
Since we need a common currency globally set for a store, hence the currency is fetched from the store settings at every place where currency is used.

## Describe your changes

1. Added a list of currencies to choose from in general settings form.
2. Used the currency set in general config everywhere as default currency.

------

## Checklist

------
Dear Gatekeeper,


apps/admin_app/lib/admin_app_web/templates/general_settings/_form.html.eex
```diff
@@ -22,7 +22,15 @@
                 <%= input f, :hosted_payment_url, nil, is_horizontal: true, description: "Payment URL for the store."  %>
             </div>
             <div class="form-group row ">
-                <%= input f, :currency, nil, is_horizontal: true, description: "Default currency for the store."  %>
+                <label  class="col-sm-3 col-form-label">
```
pkrawat1: indentation is not correct

apps/snitch_core/lib/core/data/model/general_configuration.ex
```diff
@@ -10,6 +10,12 @@ defmodule Snitch.Data.Model.GeneralConfiguration do
   alias Snitch.Tools.Helper.ImageUploader
   alias Ecto.Multi
 
+  @spec fetch_currency() :: String.t()
+  def fetch_currency do
+    general_config = Repo.all(GC) |> List.first()
+    if general_config != nil, do: general_config.currency, else: "USD"
+  end
+
```
pkrawat1:
```elixir
  def fetch_currency do
    case Repo.one(GC)
      nil -> "USD"
      gc -> gc.currency
    end
  end
```

apps/snitch_core/lib/core/domain/order/order.ex
```diff
@@ -38,7 +38,7 @@ defmodule Snitch.Domain.Order do
   """
   @spec payments_total(Order.t(), String.t()) :: Money.t()
   def payments_total(order, payment_state) do
-    {:ok, currency} = Defaults.fetch(:currency)
+    currency = GCModel.fetch_currency()
```
ashish173: I would really prefer to have `{:ok, currency}` or `{:error, msg}` rather than returning just as currency. this way matching becomes easier. @SagarKarwande thoughts?

apps/admin_app/lib/admin_app_web/views/general_settings_view.ex
```diff
@@ -2,7 +2,13 @@ defmodule AdminAppWeb.GeneralSettingsView do
   use AdminAppWeb, :view
   alias Snitch.Data.Model.ProductBrand
 
+  @currencies ["USD", "INR", "GDP", "EUR"]
```
ashish173: This should go in the core rather than in the admin application IMO. @pkrawat1 @SagarKarwande ?

apps/admin_app/lib/admin_app_web/views/layout_view.ex
```diff
@@ -36,7 +37,8 @@ defmodule AdminAppWeb.LayoutView do
   end
 
   def check_general_settings() do
-    Repo.all(GeneralConfiguration) |> List.first()
+    count = Repo.aggregate(from(g in "snitch_general_configurations"), :count, :id)
+    if count == 1, do: true, else: false
```
ashish173: This method is too complex to identify the presence of GC or not, please simplify this.

apps/admin_app/lib/admin_app_web/views/general_settings_view.ex
pkrawat1: I was thinking if any other app in the umbrella would be needing the list of currencies. If it does then it will be required to move to some other place. So how about keeping it at a place in app that can expose this to all other apps.

apps/snitch_core/lib/core/data/model/general_configuration.ex
```diff
@@ -10,6 +10,12 @@ defmodule Snitch.Data.Model.GeneralConfiguration do
   alias Snitch.Tools.Helper.ImageUploader
   alias Ecto.Multi
 
+  @spec fetch_currency() :: String.t()
+  def fetch_currency do
+    general_config = Repo.all(GC) |> List.first()
```
ashish173: Why `Repo.all()` when we can use `Repo.one()`? That way you also won't have to do `List.first()`

apps/admin_app/lib/admin_app_web/views/general_settings_view.ex
SagarKarwande: I think `core` should contain list of valid currencies

apps/snitch_core/lib/core/domain/order/order.ex
pkrawat1: We can use just the return value since we will be using a `Repo.one` query, which either gives a value or nil.

apps/admin_app/lib/admin_app_web/views/general_settings_view.ex
SagarKarwande: Yes, valid currencies should be a part of core

apps/snitch_core/lib/core/data/model/general_configuration.ex
ashish173: This is looking better.

apps/snitch_core/lib/core/domain/order/order.ex
SagarKarwande: Yes returning tuples with success and error will help.

@ashish173 As per the definition of  `GeneralConfiguration.fetch_currency()` we would never be able to get the error case.

apps/snitch_core/lib/core/data/model/general_configuration.ex
jyotigautam: This is more succinct way.

apps/snitch_core/lib/core/domain/order/order.ex
ashish173: That's okay @SagarKarwande for now. Since we always start/provision a store with a currency. I think now I agree with what @pkrawat1 is saying.

jyotigautam: But there would be no {:error, msg} scenario, it will always return a string.

apps/admin_app/lib/admin_app_web/views/layout_view.ex
pkrawat1:
```elixir
def check_general_settings do
    case Repo.one(GC)
      nil -> false
      _ -> true
    end
  end
```

jyotigautam: yeah sure

apps/admin_app/lib/admin_app_web/views/general_settings_view.ex
jyotigautam: May I know the idea behind this?

ashish173:
@jyotigautam move the currency list to the core application. This is an important change.

apps/snitch_core/lib/core/data/model/general_configuration.ex
jyotigautam: yes, then I have to handle the scenario where multiple results for Repo.one() could be returned.

ashish173: `Repo.one` only returns just one result. In any case my assumption is that we can only have one general setting with the way how the code is written.

apps/admin_app/lib/admin_app_web/views/general_settings_view.ex
SagarKarwande: Any application would always depend on core and can easily access the valid currencies.

SagarKarwande:
@ashish173 @pkrawat1 
Initially during store setup General Setting is not configured until user does it.
If any product is created before setting General Settings and later on setting up general setting with different currency would break the Money calculations. Any thought about this?

Should the currency change be allowed after General Setting is created?

apps/admin_app/lib/admin_app_web/templates/product/form.html.eex
```diff
@@ -51,7 +51,7 @@
       <div class="col-sm-9 input-group">
         <%= text_input :selling_price, :amount, value: get_amount(f.data.selling_price), class: "form-control", name: "product[selling_price][amount]" %>
         <div class="input-group-append">
-          <%= select :selling_price, :currency, get_currency(),value: get_currency_value(f.data.selling_price), class: "form-control", name: "product[selling_price][currency]" %>
```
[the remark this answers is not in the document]
jyotigautam: Here the currency selected only shows one value i.e, from the general settings or "usd" if no general setting is set. So, the user can't choose from this list.

ashish173:
@SagarKarwande we are provisioning the general settings with a currency "USD" for now. So all the products which get created happen with the currency. 

Currency change is another issue in itself we will revisit it at a later phase.

apps/admin_app/lib/admin_app_web/templates/product/form.html.eex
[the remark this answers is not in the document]
ashish173: @jyotigautam @SagarKarwande I would say we remove the select altogether and just show a label of the currency. Just like it is done with etsy store. Thoughts?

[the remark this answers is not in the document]
jyotigautam: Yes, there is no point of showing it as select input when we have just one option.

PR #339, opened 2018-12-03
Set main image for a product

jyotigautam:
Adding checkboxes alongside images to set the main image for the products. 

## Motivation and Context
Presently the images added to a product have no order, hence there should be a provision to select an image and make it as a default image to be shown for a particular product.

## Describe your changes
Added a checkbox on the images to select the main image to show as front image of the product.

Resolves [162156592](https://www.pivotaltracker.com/story/show/162156592)

------
## Checklist

------
Dear Gatekeeper,


apps/admin_app/assets/js/views/product/edit.js
```diff
@@ -31,6 +32,25 @@ export function imageOnEnter(){
   });
 }
 
+export function defaultImage(){
```
ashish173: @jyotigautam shouldn't this method have a more meaningful name, based on what it does. I see that this method sets an image as default image. Based on this I would name the method as `setDefaultImage` or `updateDefaultImage` let me know what you think?

apps/admin_app/assets/js/views/product/edit.js
```diff
@@ -31,6 +32,25 @@ export function imageOnEnter(){
...
+  $(document).delegate('input[type="checkbox"]', 'click', function(event) {
+    $(this).parent().siblings().find(':checkbox').prop('checked', false);
+    var default_image = $(this).val();
+    var product_id = $("#product-id").val();
+    var data = { product_id: product_id,  default_image: default_image};
+    var CSRF_TOKEN = $("meta[name='csrf-token']").attr("content");
```
ashish173: @jyotigautam Can we please leave an empty line in between the batch of instructions such as the above is initialisation part where we are settings variable values and below this line we are sending an ajax request? What do you think?

apps/admin_app/assets/js/views/product/edit.js
```diff
@@ -31,6 +32,25 @@ export function imageOnEnter(){
...
+      cache: false,
+      data: data,
+      beforeSend: function (xhr) {
+        xhr.setRequestHeader("X-CSRF-Token", CSRF_TOKEN);
+      }
+    });
+ });
+}
+
 export function imageOnSubmit(){
```
ashish173: @jyotigautam The function name is clear to describe the working of the function. This can be better.

apps/admin_app/assets/js/views/product/edit.js
jyotigautam: Yes, defaultImage() doesn't make it clear what exactly is done with the image. You are right, fixing this in the next commit.

ashish173:
@jyotigautam I was thinking when we upload the images the first image should automatically be set as the main image. What do you think?

apps/admin_app/assets/js/views/product/edit.js
jyotigautam: imageOnSubmit function is not required now since we have removed the submit button from the upload image page.

jyotigautam:
> @jyotigautam I was thinking when we upload the images the first image should automatically be set as the main image. What do you think?
> 
> Also, please update the test cases. The coverage is negative.

Ideally, yes.

apps/snitch_api/lib/snitch_api_web/views/product_view.ex
```diff
@@ -37,6 +38,14 @@ defmodule SnitchApiWeb.ProductView do
     product.selling_price |> to_string
   end
 
+  def default_image(product, _conn) do
+    product = product |> Repo.preload(:images)
```
SagarKarwande: Can the default image be preloaded directly using query instead of preloading all images.

apps/snitch_api/lib/snitch_api_web/views/product_view.ex
```diff
@@ -37,6 +38,14 @@ defmodule SnitchApiWeb.ProductView do
...
+
+    product.images
+    |> Enum.filter(fn image -> image.is_default == true end)
+    |> Enum.map(fn image -> %{"default_product_url" => Product.image_url(image.name, product)} end)
```
SagarKarwande: `Default image` key will have array value type, as we are using Enum.map

apps/snitch_api/lib/snitch_api_web/views/product_view.ex
jyotigautam: Ohh yes, conditional preloading for images can be done in this case.

apps/admin_app/lib/admin_app_web/controllers/product_controller.ex
```diff
@@ -88,21 +89,58 @@ defmodule AdminAppWeb.ProductController do
...
-  def add_images(conn, %{"product_images" => product_images, "product_id" => id}) do
-    product =
-      id
-      |> String.to_integer()
-      |> ProductModel.get()
-      |> Repo.preload(:images)
+  defp preload_product_images(id) do
+    id
+    |> String.to_integer()
+    |> ProductModel.get()
```
SagarKarwande: No need to convert to string id integer as ProductModel support string.
`Snitch.Tools.Helper.Query` module has been update to use string ids.

apps/snitch_api/lib/snitch_api_web/views/product_view.ex
```diff
@@ -37,6 +41,17 @@ defmodule SnitchApiWeb.ProductView do
     product.selling_price |> to_string
   end
 
+  def default_image(product, _conn) do
+    default_image = from(image in Image, where: image.is_default == true)
+
+    query =
+      from(p in ProductSchema, where: p.id == ^product.id, preload: [images: ^default_image])
+
+    product = Repo.one(query)
```
SagarKarwande: Handle clause where we are not able to fetch the default image.

Steps to reproduce:
1. Add images to any `active` product. 
2. Delete the default image.
3. Try to call the product listing API.

API will fail.

ashish173:
don't allow user to delete the default image, if a user wants to delete the default image they have to change the default image to some other image.

PR #342, opened 2018-12-03
Fix long URL issue after product is edited with long description

vikramjadon921995:
This change will solve the issue of server returning 414 status code when very large description is added in the product edit page.

[Fixes Pivotal - 162307066](https://www.pivotaltracker.com/story/show/162307066)

## Motivation and Context
When we try to edit a product with long text description and then save the product.
The data for product gets saved but the product fields are present in URI after saving the data. As the description is too long the URL also becomes long and we get 414 status code on the page after redirect.

## Describe your changes
Sending default rummage instead of data from the product description form field in update product action.

------

## Checklist

------
Dear Gatekeeper,

### DIFF
--- a/apps/admin_app/lib/admin_app_web/controllers/product_controller.ex
+++ b/apps/admin_app/lib/admin_app_web/controllers/product_controller.ex
@@ -75,7 +75,7 @@ defmodule AdminAppWeb.ProductController do
   def update(conn, %{"product" => params}) do
     with %ProductSchema{} = product <- ProductModel.get(params["id"]),
          {:ok, _product} <- ProductModel.update(product, params) do
-      redirect_with_updated_conn(conn, params)
+      redirect_with_updated_conn(conn, @rummage_default)
     end
   end
 


apps/admin_app/lib/admin_app_web/controllers/product_controller.ex
```diff
@@ -75,7 +75,7 @@ defmodule AdminAppWeb.ProductController do
   def update(conn, %{"product" => params}) do
     with %ProductSchema{} = product <- ProductModel.get(params["id"]),
          {:ok, _product} <- ProductModel.update(product, params) do
-      redirect_with_updated_conn(conn, params)
+      redirect_with_updated_conn(conn, @rummage_default)
```
anantanant2015: rummage_default is one solution, but I want the functionality if the user has done edit, with rummage opts, sort as query params, then those needs to be preserved, which in turn will do the redirection with same filters as were before not the defaults.

PR #343, opened 2018-12-04
Append schema in definitions of exclusivity functions

ramansah:
## Motivation and Context
Exclusivity functions defined in database mentioned tenant specific table in their respective definitions . Since these functions were defined in public schema, existing functions were being replaced on creating a new tenant and throw errors when used by the former tenant.

## Describe your changes
Schema prefixes are added in the definition of DB exclusivity functions.
Therefore, instead of using public.payment_exclusivity, 
tenant.payment_exclusivity will get invoked.

## Any further comments?

------

## Checklist

------
Dear Gatekeeper,

### DIFF
--- a/apps/snitch_core/priv/repo/migrations/20180222125936_create_payment_tables.exs
+++ b/apps/snitch_core/priv/repo/migrations/20180222125936_create_payment_tables.exs
@@ -2,7 +2,7 @@ defmodule Core.Repo.Migrations.CreatePaymentTables do
   use Ecto.Migration
 
   @payment_exclusivity_fn ~s"""
-  create or replace function payment_exclusivity(
+  create or replace function #{ prefix() || "public" }.payment_exclusivity(
     in supertype_id bigint,
     in subtype_discriminator char(3)
     )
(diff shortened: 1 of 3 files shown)


SagarKarwande:
There are total three functions: `money_state_function`, `payment_exclusivity`, `zone_exclusivity`.
`payment_exclusivity` and `zone_exclusivity` are already fixed in this PR.

Just want to know if `money_state_function` needs to be created in all schemas.

@anantanant2015 I did not find any use of `money_state_function`. Where do we need this? Do we need this in each schema?

ramansah:
I don't think `money_state_function` needs to be created for all schemas since it acts on `money_with_currency` and that is global for all schemas (defined in public).
Logically too, money, currency types and related functions don't change in different tenants and it would be advisable to keep them in public schema.

anantanant2015:
> There are total three functions: `money_state_function`, `payment_exclusivity`, `zone_exclusivity`.
> `payment_exclusivity` and `zone_exclusivity` are already fixed in this PR.
> 
> Just want to know if `money_state_function` needs to be created in all schemas.
> 
> @anantanant2015 I did not find any use of `money_state_function`. Where do we need this? Do we need this in each schema?

I did not wrote these functions they have generated after running a script.

PR #354, opened 2018-12-10
Order states using enum

anantanant2015:
This would update order to use states as enum.

## Why?

- Order states should use predefined enum states, and should not
  use strings to define state for any order.

- Adds states for order, and updates order states as enum.

[delivers #162373423]

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,21 +1,12 @@
-<!--- Provide a general summary of your changes in the Title above -->
-<!--- in NOT MORE THAN 50 characters. Keep it short, pls. -->
-
-## Motivation and Context
+## Why?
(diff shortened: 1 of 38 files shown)


apps/snitch_core/mix.exs
```diff
@@ -60,7 +60,7 @@ defmodule Snitch.Core.Mixfile do
       {:ecto_autoslug_field, "~> 0.5"},
 
       # state machine
-      {:beepbop, github: "oyeb/beepbop", branch: "develop"},
+      {:beepbop, github: "anantanant2015/atom_beepbop", branch: "develop"},
```
ashish173: @anantanant2015  Why `atom_beepbop`, just keep it `beepbop` would be enough imo.

anantanant2015: done

PR #358, opened 2018-12-12
Display payment method name in order details.

anantanant2015:
## Why?

- We want to display payment name in order details page, instead of showing
  payment method code.

- Because to make order details page more complete to give user a better
  understanding for payment method.

[delivers #162570811]

## Checklist


apps/admin_app/lib/admin_app_web/views/order_view.ex
```diff
@@ -272,4 +273,9 @@ defmodule AdminAppWeb.OrderView do
   def get_docs_url() do
     Application.get_env(:admin_app, AdminAppWeb.Endpoint)[:docs_url]
   end
+
+  def get_payment_name(payment) do
+    payment = payment |> Repo.preload(:payment_method)
```
arjun289: Instead of making a query here, check the context from where data is being loaded while showing order detail and preload there itself.

PR #376, opened 2018-12-21
Product export

jyotigautam:
## Why?
To generate csv/xlsx data export for orders.

1. Generating the data for products in csv/xlsx formats.
2.  Implementing a worker to send the generated file to user via mail as a background task.

[delivers #162571469](https://www.pivotaltracker.com/story/show/162571469)

## Checklist


apps/admin_app/lib/admin_app_web/helpers.ex
```diff
@@ -122,30 +158,35 @@ defmodule AdminAppWeb.Helpers do
...
-    columns =
-      ~w(id number line_items_count order_total billing_address shipping_address inserted_at updated_at user_id state)
+  def xlsx_generator(data_list, type) do
+    columns = get_columns(type) |> Enum.map(&Atom.to_string(&1))
 
-    orders = orders |> Enum.map(&parse_line(&1))
-    rows = orders |> Enum.map(&row(&1))
-    %Workbook{sheets: [%Sheet{name: "Orders", rows: [columns] ++ rows}]}
+    data_list = data_list |> Enum.map(&parse_line(&1))
+    rows = data_list |> Enum.map(&row(&1, columns))
```
ashish173: IMO this can written simply as:-

```
rows = data_list 
|> Enum.map(&parse_line(&1))
|> Enum.map(&row(&1, columns))
```

what do you think? @jyotigautam

apps/admin_app/lib/admin_app_web/helpers.ex
```diff
@@ -122,30 +158,35 @@ defmodule AdminAppWeb.Helpers do
     end
   end
 
-  def order_xlsx_exporter(user) do
-    orders = Repo.all(Order) |> Repo.preload([:line_items])
+  def xlsx_exporter(user, type) do
```
ashish173: @jyotigautam this should definitely be moved out of helpers. @SagarKarwande @pkrawat1 what do you guys think?

apps/admin_app/lib/admin_app_web/helpers.ex
```diff
@@ -122,30 +158,35 @@ defmodule AdminAppWeb.Helpers do
...
+    data_list =
+      case type do
+        "order" ->
+          Repo.all(Order) |> Repo.preload([:line_items])
```
ashish173: Moreover, these methods to get all the orders and products should go in product models not here. We have to expose a method from domain and use them here in the application. 

Think of it as this way. If we ever need to write a custom logic in the core app which accesses this logic then we have to certainly write this logic again. there is no way we'll be able to reuse this custom order/product logic to get all products and orders. 

Please pay more attention to DRY and SRP when you are coding. This is and will be violating both the principles.

apps/admin_app/lib/admin_app_web/helpers.ex
```diff
@@ -122,30 +158,35 @@ defmodule AdminAppWeb.Helpers do
...
+
+        "product" ->
+          Repo.all(Product)
+          |> Repo.preload([:variants, :taxon, :shipping_category, :options, :brand, :theme])
```
ashish173: same here again.

apps/admin_app/lib/admin_app_web/router.ex
```diff
@@ -31,6 +31,9 @@ defmodule AdminAppWeb.Router do
     # Use the default browser stack
     pipe_through([:browser, :authentication])
 
+    get("/", PageController, :index)
```
ashish173: @jyotigautam can you give a brief explanation about this for me? Thanks!

apps/admin_app/lib/admin_app_web/helpers.ex
ashish173: Please see [here](https://github.com/spree/spree/tree/master/core/lib/spree/core/importer) how it has been organised in the `lib` folder. We need to do like this. I don't think we have something like this as of now.

apps/admin_app/lib/admin_app_web/views/product_view.ex
```diff
@@ -42,7 +42,7 @@ defmodule AdminAppWeb.ProductView do
     length(product.taxon.variation_themes) > 0
   end
 
-  defp is_child_product(product) do
+  def is_child_product(product) do
```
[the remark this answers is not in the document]
SagarKarwande: We can move this to product model

apps/admin_app/lib/admin_app_web/router.ex
SagarKarwande: @jyotigautam Can we remove this route as this route has been moved to [line 110](https://github.com/aviacommerce/avia/blob/c5c274762c73caa6e16d7a8e4259240cab237d0a/apps/admin_app/lib/admin_app_web/router.ex#L110) in [PR 375 - React integration](https://github.com/aviacommerce/avia/pull/375).

This will break the react integration and the request will not be serverd properly by react.

apps/admin_app/lib/admin_app_web/mailer/data_export_mail.ex
```diff
@@ -1,6 +1,6 @@
 defmodule AdminAppWeb.DataExportMail do
   @moduledoc """
-  Composing email to send order export.
+  Composing email to send order/product data export.
```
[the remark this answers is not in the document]
jyotigautam: I kept it like this considering the current requirement, but yes this should be generic to send all kind of attachments.

apps/admin_app/lib/admin_app_web/helpers.ex
jyotigautam: Refactoring it this way.

apps/admin_app/lib/admin_app_web/views/product_view.ex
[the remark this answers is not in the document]
jyotigautam: I would suggest to move is_parent_product function as well.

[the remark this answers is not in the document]
ashish173: yes, these are all product model related functions. @jyotigautam thanks for your suggestion 👍

apps/snitch_core/lib/core/data/model/product.ex
```diff
@@ -315,6 +323,26 @@ defmodule Snitch.Data.Model.Product do
     end)
   end
 
+  # TODO This needs to be replaced and we need a better system to identify
+  # the type of product.
```
ashish173: @jyotigautam add specs for this method please.

ashish173: add specs for any other functions in this module without specs.

apps/admin_app/lib/admin_app_web/templates/product/form.html.eex
```diff
@@ -188,7 +188,7 @@
       <% end %>
     </div>
    <% else %>
-    <%= if !is_child_product(@parent_product) do %>
+    <%= if !Product.is_child_product(@parent_product) do %>
```
ashish173: @jyotigautam can we use `unless` here instead of `if !Product...` that would more succinct.

apps/admin_app/lib/admin_app_web/templates/product/form.html.eex
```diff
@@ -10,7 +10,7 @@
       <a class="nav-link active" href="#tab1default" data-toggle="tab">Product Detail</a>
    </li>
    <%= if  @conn.request_path != product_path(@conn, :new) do %>
-   <%= if !is_child_product(@parent_product) do %>
+   <%= if !Product.is_child_product(@parent_product) do %>
```
ashish173: @jyotigautam use `unless` here please.

apps/admin_app/lib/admin_app_web/exporter/exporter.ex
```diff
@@ -0,0 +1,93 @@
...
+  alias Snitch.Domain.Order, as: Domain
+  alias Snitch.Core.Tools.MultiTenancy.Repo
+  alias Elixlsx.{Workbook, Sheet}
+  alias Snitch.Data.Schema.{Order, Product}
+  alias AdminAppWeb.DataExportMail
+
+  defp get_columns(type) do
+    case type do
+      "order" ->
+        ~w(id number line_items_count order_total billing_address shipping_address inserted_at updated_at user_id state)a
```
ashish173: @jyotigautam I see this as repetition from [line](https://github.com/aviacommerce/avia/pull/376/files#diff-1fe76619644f2dc3a5d5e02bdc3ebb5eR11).

Same goes for line 20 for production options. 

This violates the [DRY](https://metova.com/dry-programming-practices/) principle. 

Am I missing something here? Please propose a solution for this if my understanding is correct.

apps/admin_app/lib/admin_app_web/templates/product/form.html.eex
jyotigautam: thanks for this.

apps/admin_app/lib/admin_app_web/templates/product/form.html.eex
pkrawat1: @jyotigautam @ashish173 
I would say call it from helper. Calling module name in views will make it difficult to change.

Like what if someone wants to rename `is_child_product` later. Or want to modify its `arity` later. Then all the views will be needed to be updated.

Instead just call the function from a view helper.

apps/admin_app/lib/admin_app_web/exporter/exporter.ex
jyotigautam: I forgot to remove this. I have passed column list for product and order as arguments from product exporter and order exporter function respectively.

apps/admin_app/lib/admin_app/honeydew_workers/export_data_worker.ex
```diff
@@ -0,0 +1,18 @@
...
+    Repo.set_tenant(tenant)
+
+    case format do
+      "csv" ->
+        Exporter.csv_exporter(user, type)
+
+      "xlsx" ->
+        Exporter.xlsx_exporter(user, type)
+    end
+  end
```
pkrawat1: @ashish173  @SagarKarwande @arjun289 @jyotigautam @anantanant2015 @vikramjadon921995 @ramansah 
What might be wrong in this approach?
I think its related to how atoms behave in elixir and they are limited.

```elixir
  def export_data(%{"tenant" => tenant, "format" => format, "type" => type, "user" => user}) do
    Repo.set_tenant(tenant)
    apply(Exporter, :"#{format}_exporter", [user, type])
  end
```

`RAILS`
- in rails we do `send` with string params. There it should not be an issue. since not using symbols.

apps/admin_app/lib/admin_app_web/helpers.ex
```diff
@@ -1,12 +1,8 @@
...
-  alias Ecto.Adapters.SQL
+
   alias Snitch.Core.Tools.MultiTenancy.Repo
-  alias Snitch.Data.Schema.Order
-  alias Snitch.Domain.Order, as: Domain
-  alias Elixlsx.{Workbook, Sheet}
-  alias AdminAppWeb.DataExportMail
+  alias Snitch.Data.Schema.{Order, Product}
 
   @months ["Jan", "Feb", "Mar", "Apr", "May", "June", "July", "Aug", "Sept", "Oct", "Nov", "Dec"]
```
ashish173: Do we really need months like these? @jyotigautam can't it be somehow be fetched from `Calendar` module or something. Names of months is a pretty generic thing to define as a module attribute in my opinion.

apps/snitch_core/lib/core/data/model/order.ex
```diff
@@ -201,9 +201,15 @@ defmodule Snitch.Data.Model.Order do
...
   @doc """
-    Order related to user.
+  Returns all Orders with the given list of entities preloaded
   """
+  def get_all_with_preloads(preloads) do
+    Repo.all(Order) |> Repo.preload(preloads)
+  end
 
+  @doc """
+    Order related to user.
```
ashish173: In correct indentation for the comment 👎

apps/admin_app/lib/admin_app_web/helpers.ex
jyotigautam: No, we need to  have a list of month names since the Calendar module or any other module dealing with date time returns month as a numerical value(1-12)

vikramjadon921995: We cab use strftime: https://hexdocs.pm/timex/Timex.Format.DateTime.Formatters.Strftime.html
What is the usecase? Do we need to format full date and time? Or find name of month for a specific month number ?

vikramjadon921995: We can format whole date using strftime. Need to check how to get month name using this: https://hexdocs.pm/timex/Timex.Format.DateTime.Formatters.Strftime.html#module-months

suhasdombale123: sir i found out this ...Elixir.Timex.month_shortname(1)

ashish173: @suhasdombale123 I appreciate your proactive response! Keep it up.

ashish173:
@jyotigautam PR is decreasing coverage. Please fix it.

ashish173:
@jyotigautam `sellable_products_query` and `get_all_with_preloads` methods are not tested. These are important methods which should be tested to avoid any future regressions.

apps/snitch_core/lib/core/data/model/product.ex
```diff
@@ -28,7 +28,11 @@ defmodule Snitch.Data.Model.Product do
   """
   @spec get_all_with_preloads(list) :: [Product.t()]
   def get_all_with_preloads(preloads) do
-    Repo.all(Product) |> Repo.preload(preloads)
+    try do
+      Repo.all(Product) |> Repo.preload(preloads)
+    rescue
+      e in ArgumentError -> nil
+    end
```
pkrawat1: @jyotigautam May I know the reason to use `rescue` block here

pkrawat1: We are in control of what params it gets right?

PR #377, opened 2018-12-21
Fixed release task to automatically run migrations for each tenant

ramansah:
Running migrations on tenants failed in production since Mix is a build tool and [not available in production releases](https://elixirforum.com/t/mix-project-missing-in-elixir-phoenix-prod-release/11413). 
Earlier implementation relied on Mix to run migrations.
The fix runs migrations directly through Ecto's methods.

### DIFF
--- a/apps/snitch_core/priv/tasks/release_tasks.ex
+++ b/apps/snitch_core/priv/tasks/release_tasks.ex
@@ -16,7 +16,7 @@ defmodule Snitch.Tasks.ReleaseTasks do
     stop_services()
   end
 
-  def seed()do
+  def seed() do
     start_services()
 
     run_migrations()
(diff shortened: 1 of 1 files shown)


ashish173:
@ramansah @pkrawat1 If my understanding is correct this will take a lot of time to run. I am just wondering Is there any other way to achieve this that we can think of?

ramansah:
> @ramansah @pkrawat1 If my understanding is correct this will take a lot of time to run. I am just wondering Is there any other way to achieve this that we can think of?

Ideally this is very fast but it may take a while in two cases.
1. Very large number of tenants
2. VLDB - modifying tables is a costly operation.
Improvement in case 1 is possible by running migrations on tenants in parallel.
Case 2 requires some additional thinking since the app may throw errors if migrations are not complete and modified tables are accessed from the new version (assuming Honeydew is used).

BTW the migrations tasks in distillery are under `post_start` hook so the app gets started before running migrations.

ashish173:
> Improvement in case 1 is possible by running migrations on tenants in parallel.

What might be issues with this approach. I am not aware of this scenario.

> Case 2 requires some additional thinking since the app may throw errors if migrations are not complete and modified tables are accessed from the new version (assuming Honeydew is used).

Yes, this is a common issue with migrations and deployments. 

> BTW the migrations tasks in distillery are under post_start hook so the app gets started before running migrations.

Does this not raise the same issues as the second case. Let's say for example if we have introduced new fields?

In any case, I think we are good for now. We'll continue to monitor this and looks for other approaches when it's starts to become a bottleneck. 

Thanks for the PR 🥇

PR #381, opened 2018-12-24
Package enum state

anantanant2015:
## Why?

- Package states should use predefined enum states, and should not use strings to define state for
  any Package.

- Adds states for Package, and updates package states as enum.

[delivers #162636456]

## Checklist


apps/snitch_core/test/data/schema/package_test.exs
```diff
@@ -57,7 +57,7 @@ defmodule Snitch.Data.Schema.PackageTest do
   @update_params %{
     shipping_methods: [],
     tracking: %{id: "some_tracking_id"},
-    state: "ready!",
+    state: "ready",
```
ashish173: @anantanant2015 shouldn't this be atom? if not why not?

anantanant2015: because when params are casted then they can be either in type integer, string or atom, since in the controller params are received in string form, here are the valid values map
PackageStateEnum.__valid_values__()
[0, 1, 2, 3, 4, 5, 6, :pending, :processing, :fulfilled, :ready, :shipped,
 :delivered, :complete, "complete", "delivered", "fulfilled", "pending",
 "processing", "ready", "shipped"]

PR #389, opened 2018-12-27
Variant state change

jyotigautam:
## Why?
To change variant state on the product edit page of the admin panel.

Providing toggle switch alongside each variant to change it's state.

[delivers #162779292](https://www.pivotaltracker.com/story/show/162779292)

## Checklist


apps/admin_app/assets/js/views/product/edit.js
```diff
@@ -35,6 +36,23 @@ export function imageOnEnter() {
   });
 }
 
+export function onToggleActivation() {
+  $(document).delegate('.variant_activate', 'change', function (event) {
```
SagarKarwande: `delegate` has been deprecated, try to use alternative for this.
Follow the docs in this [link](http://api.jquery.com/delegate/)

apps/admin_app/assets/js/views/product/edit.js
```diff
@@ -13,6 +13,7 @@ export default class View extends MainView {
     handleImageSelect();
     deleteImage();
     setupProduct();
+    onToggleActivation();
```
SagarKarwande: Method name does not give the intent of its action

apps/admin_app/assets/js/views/product/edit.js
```diff
@@ -35,6 +36,23 @@ export function imageOnEnter() {
...
 
+export function onToggleActivation() {
+  $(document).delegate('.variant_activate', 'change', function (event) {
+    var id = $(this).find('.variant_id').val();
+    var target_element = $(this);
+    var state = ( $(this).find('.switch-input').prop("checked") == true) ? "active" : "draft";
+    var data = {state: state, id: id};
+    $.ajax({
+      url: `/variant_state`,
+      type: "GET",
```
SagarKarwande: Should the method request be `PATCH`?

apps/admin_app/lib/admin_app_web/controllers/product_controller.ex
```diff
@@ -112,6 +112,15 @@ defmodule AdminAppWeb.ProductController do
     redirect_with_updated_conn(conn, params)
   end
 
+  def toggle_variant_state(conn, %{"state" => state, "id" => product_id}) do
```
SagarKarwande: We are getting state from frontend it might be possible that the state can be invalid.
Porduct update will fail with changeset. That clause needs to be handled.

apps/admin_app/lib/admin_app_web/templates/product/form.html.eex
```diff
@@ -167,16 +167,30 @@
...
           </thead>
           <tbody>
             <%= for variant <- @parent_product.variants do %>
-            <tr>
+            <tr class="product_variants">
               <td><%= variant.id %></td>
               <%= for option <- variant.options do %>
               <td class="option-value" contenteditable="true" data-option_value_id="<%= option.id %>"><%= option.value %></td>
               <% end %>
+              <td class="variant_state"><%= variant.state %></td>
```
SagarKarwande: We are using the content of td tag as state for the update. I think using it from `data attribute` would be proper.

apps/admin_app/assets/js/views/product/edit.js
jyotigautam: Ideally yes, if it deals with updating just specific field in a struct.

apps/admin_app/lib/admin_app_web/templates/product/form.html.eex
SagarKarwande: Ok so `variant_state` is used as a selector to update the state in HTML. LGTM

PR #423, opened 2019-01-15
 Test cases for country model

suhasdombale123:
## Why?
To ensure that all the functions in country model are tested 

added tests in model/country_test.exs file

[delivers #163144268](https://www.pivotaltracker.com/story/show/163144268)

## Checklist

### DIFF
--- a/apps/snitch_core/test/data/model/country_test.exs
+++ b/apps/snitch_core/test/data/model/country_test.exs
@@ -6,8 +6,8 @@ defmodule Snitch.Data.Model.CountryTest do
 
   setup :countries
 
-  describe "Country formatting" do
-    test "for a single country ", %{countries: countries} do
+  describe "formatted_list/0" do
+    test "succeeds ", %{countries: countries} do
       country = countries |> List.first()
       country_list = Country.formatted_list()
       assert country_list == [{country.name, country.id}]
--- a/apps/snitch_core/test/data/model/country_test.exs
+++ b/apps/snitch_core/test/data/model/country_test.exs
@@ -17,7 +17,6 @@ defmodule Snitch.Data.Model.CountryTest do
   describe "get/1" do
     test "succeeds", %{countries: countries} do
       country = countries |> List.first()
-
       country_list = Country.get(%{numcode: country.numcode})
       assert country_list == country
     end
--- a/apps/snitch_core/test/data/model/country_test.exs
+++ b/apps/snitch_core/test/data/model/country_test.exs
@@ -1,0 +1,37 @@
+defmodule Snitch.Data.Model.CountryTest do
+  use ExUnit.Case, async: true
+  use Snitch.DataCase
+  alias Snitch.Data.Model.Country
+  import Snitch.Factory
+
+  setup :countries
+
+  describe "Country formatting" do
+    test "for a single country ", %{countries: countries} do
+      country = countries |> List.first()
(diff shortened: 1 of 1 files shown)


apps/snitch_core/test/data/model/country_test.ex
```diff
@@ -0,0 +1,39 @@
...
+      country = countries |> List.first()
+
+      country_list = Country.get(%{numcode: country.numcode})
+      assert country_list == country
+    end
+
+    test "fails, Why? value not present for a given key", %{countries: countries} do
+      country = countries |> List.first()
+
+      country_list = Country.get(%{numcode: "842"})
```
SagarKarwande: We can test without fetching the first country from the list and with the actual value

```elixir
assert Country.get(%{numcode: "842"}) == nil
```

apps/snitch_core/test/data/model/country_test.ex
```diff
@@ -0,0 +1,39 @@
...
+      refute country_list == country
+    end
+
+    test "for getting all records ", %{countries: countries} do
+      country = countries |> List.first()
+
+      country_list = Country.get_all()
+      assert country_list|> List.first() == country
```
SagarKarwande: Can we assert the length instead of the first element.

SagarKarwande (changes requested):
Run elixir formatter.

apps/snitch_core/test/data/model/country_test.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Snitch.Data.Model.CountryTest do
+  use ExUnit.Case, async: true
+  use Snitch.DataCase
+  alias Snitch.Data.Model.Country
+  import Snitch.Factory
+
+  setup :countries
+
+  describe "Countries" do
```
arjun289: `describe` is usually used to group tests related to a particular type. 
e.g. It would be better to group tests by `functions` in the model. Here single describe has no purpose since all the tests are within this.

apps/snitch_core/test/data/model/country_test.ex
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    test "for getting records by a particular key ", %{countries: countries} do
+      country = countries |> List.first()
+
+      country_list = Country.get(%{numcode: country.numcode})
+      assert country_list == country
+    end
+
+    test "fails, Why? value not present for a given key", %{countries: countries} do
```
arjun289: If you look at test case calls, the statement generated won't make any sense. What fails here? If you group it by function name 
```
describe `get/1` do
   test "fails" do
   end

   test "succeeds" do
   end
end
```

This would make for more meaningful statements.

apps/snitch_core/test/data/model/country_test.ex
```diff
@@ -0,0 +1,39 @@
...
+      country = countries |> List.first()
+
+      country_list = Country.get(%{numcode: "842"})
+      refute country_list == country
+    end
+
+    test "for getting all records ", %{countries: countries} do
+      country = countries |> List.first()
```
arjun289: `List.first()` may fail since the order is not defined in `get_all`

arjun289 (changes requested):
Try and group together tests in `describe` for different functions, it will also allow grouping edge cases better.

SagarKarwande (approved):

PR #425, opened 2019-01-15
Add upn number to uniquely identify products

jyotigautam:
## Why?
Instead of using just slug to uniquely identify products, we are using upn (unique product number) analogous to amazon's asin number.

1. Use unique number(upn) for products instead of slug to uniquely identify them.
2. Added upn field to product schema.

Migration added:
20190115093617_add_upn_field_to_products.exs

[finishes  #163114687](https://www.pivotaltracker.com/story/show/163114687)

## Checklist

### DIFF
--- a/apps/snitch_core/lib/core/data/model/product.ex
+++ b/apps/snitch_core/lib/core/data/model/product.ex
@@ -513,6 +513,19 @@ defmodule Snitch.Data.Model.Product do
   end
 
   def upn_generate() do
-    "A" <> Nanoid.generate(10, "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ") <> "C"
+    upn = "A" <> Nanoid.generate(10, "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ") <> "C"
+
+    case is_upn_unique(upn) do
+      true ->
+        upn
+
+      false ->
+        upn_generate()
+    end
+  end
+
+  defp is_upn_unique(upn) do
+    is_present = from(p in "snitch_products", select: p.upn, where: p.upn == ^upn) |> Repo.one()
+    if is_present == nil, do: true, else: false
   end
 end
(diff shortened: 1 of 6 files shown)


apps/snitch_core/lib/core/data/model/product.ex
```diff
@@ -511,4 +511,8 @@ defmodule Snitch.Data.Model.Product do
   def is_variant_tracking_enabled?(product) do
     Product.is_variant_tracking_enabled?(product)
   end
+
+  def upn_generate() do
+    "A" <> Nanoid.generate(10, "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ") <> "C"
```
ashish173: @jyotigautam I thought we only agreed on numbers in between the `A` & `C`. Did I miss something?

apps/snitch_core/priv/repo/migrations/20190115093617_add_upn_field_to_products.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Snitch.Repo.Migrations.AddUpnFieldToProducts do
+  use Ecto.Migration
+
+  def change do
+    alter table("snitch_products") do
+      add(:upn, :string)
+    end
+
+    drop unique_index("snitch_products", [:slug])
```
ashish173: @jyotigautam you are dropping the slug here. But, IMO I think this should be done where we remove the product slug to fetch the product. Either make the change here for product search with `UPN`. or remove this drop index here.

ashish173:
@jyotigautam we should also make the change for the product factory in test cases. This should have major changes across the application.

apps/snitch_core/lib/core/data/model/product.ex
ashish173: change the logic here.

ashish173: I think we should also do loop check for the collision of generated `upn`.

ashish173 (approved):

PR #440, opened 2019-01-23
Add Tax Configuration and Tax Class

arjun289:
## Why?
Tax configuration is required to store general configuration data for
handling taxes. Tax Classes are used around the system as templates
which would be set for a tax rate and would assist in tax calculation.

__Tax Configration__
It has been handled using `single row` modelling wherein data is stored in single row with all the attributes of configurations as columns. 
An EAV approach was taken earlier to handle this but it has some limitations which are mentioned in 
detail [here](https://www.pivotaltracker.com/story/show/163364131).

__Note__
The older method of handling tax through `tax_categories` and `tax_rates` has been removed in this PR.

- Adding schema, migration for tax class and tax configuration.
- Adding CRUD API for the above entities.
- Adding seed for setting up initial configuration data.

[delivers #163283435]
[delivers #163282965]

## Checklist


apps/snitch_core/lib/core/data/schema/tax/tax_config.ex
```diff
@@ -0,0 +1,73 @@
...
+    to calculate taxes if prices are inclusive of tax.
+  - `default_state`: The field is used to set the default tax state. If set a zone containing
+    the state would be used for calculating tax if they are included in prices.
+  - `preferences`: A json field to store all the other params in a jsonb map.
+  """
+  @type t :: %__MODULE__{}
+
+  schema "snitch_tax_configuration" do
+    field(:label, :string)
+    field(:included_in_price?, :boolean, default: false)
```
ashish173: @arjun289 let's have `included_in_price?` default to `true`. The behaviour we want is similar to how amazon does it. Additionally, this approach does not surprise the end customer as the total cart value is exactly the sum of the cost of products. This changes drastically in the current approach here which we have taken here.  

Let me know your thoughts on this?

ashish173 (approved):

PR #451, opened 2019-01-28
User deletion

jyotigautam:
## Why?
To allow user deletion in the system.

1. Adding state (active or deleted) and deleted_at field to user and allowing soft deletion of user.
2. Add unique index to user email and deleted_at field in user schema.

Migration added:
 apps/snitch_core/priv/repo/migrations/20190128081011_update_user_unique_index.exs

Referred blog:
[Make soft deleted user email available again](https://laraveldaily.com/make-soft-deleted-user-email-available/)

[delivers #163182381](https://www.pivotaltracker.com/story/show/163182381)

## Checklist


apps/snitch_core/lib/core/data/schema/user.ex
```diff
@@ -85,23 +88,51 @@ defmodule Snitch.Data.Schema.User do
     |> common_changeset()
   end
 
-  def delete_changeset(user, params) do
+  def delete_changeset(user, _params \\ %{}) do
+    params = %{
+      "state" => "deleted",
```
ashish173: @jyotigautam we don't wanna set string states in DB. use enums.

apps/snitch_core/lib/core/data/schema/user.ex
```diff
@@ -85,23 +88,51 @@ defmodule Snitch.Data.Schema.User do
...
+      {:ok, email} ->
+        email_user_exists(email, changeset)
+
+      :error ->
+        changeset
+    end
+  end
+
+  defp email_user_exists(email, changeset) do
+    user = from(u in __MODULE__, where: u.state == "active" and u.email == ^email) |> Repo.one()
```
ashish173: @jyotigautam here also we are using strings, we should atoms? thoughts ?

apps/snitch_core/lib/core/data/schema/user.ex
```diff
@@ -85,23 +88,51 @@ defmodule Snitch.Data.Schema.User do
...
+
+    case user do
+      nil ->
+        changeset
+
+      user ->
+        changeset
+        |> add_error(:email, "email has already been taken")
```
ashish173: @jyotigautam can we please change it to "Email already in use".

apps/snitch_core/priv/repo/seed/users.ex
```diff
@@ -19,7 +19,7 @@ defmodule Snitch.Seed.Users do
       user("Harry", "Potter", "admin@aviacommerce.com", @admin_passwd, true, admin_role.id)
     ]
 
-    Repo.insert_all(User, users, on_conflict: :nothing, conflict_target: [:email])
+    Repo.insert_all(User, users, on_conflict: :nothing, conflict_target: [:email, :deleted_at])
```
ashish173, line 22: @jyotigautam need a good explanation for this.

ashish173:
@jyotigautam There are some cases we need to test for `user model`.

1. The scenario where we create -> delete a user -> create again with same email = result should be two different user records with the same email but only the active user is returned in model functions.

apps/snitch_core/priv/repo/seed/users.ex
jyotigautam: conflict target looks up for the columns where conflicts can occur, if not given explicitly, it is by default made of primary keys and/or unique constraints. Here, if I had not mentioned the conflict_target fields, then also it would have picked [:email, :deleted_at] fields according to the database constraint on user table.

apps/snitch_core/lib/core/data/schema/user.ex
jyotigautam: Enum allows us to pass an integer, atom or string to it while performing a query on the state(or any field with EctoEnum type) field.

apps/snitch_core/test/data/model/user_test.exs
```diff
@@ -83,4 +83,26 @@ defmodule Snitch.Data.Model.UserTest do
...
+    end
+
+    test "if all the users are active", %{users: users} do
+      assert length(User.get_all()) == length(users)
+    end
+
+    test "if a user is soft deleted", %{valid_attrs: va, users: users} do
+      {:ok, user} = User.create(va)
+      assert length(User.get_all()) == length(users) + 1
+      {:ok, deleted_user} = User.delete(user.id)
```
ashish173, line 101: @jyotigautam can you please leave a blank line between the grouped statements? This all looks messed up.

apps/snitch_core/test/data/schema/user_test.exs
```diff
@@ -35,6 +36,14 @@ defmodule Snitch.Data.Schema.UserTest do
       assert %{email: ["has invalid format"]} = errors_on(cs)
     end
 
+    test "if an email is already taken" do
+      user = insert(:user)
+      params = Map.update!(@valid_attrs, :email, fn _ -> user.email end)
+      cs = %{valid?: validity} = User.create_changeset(%User{}, params)
```
ashish173, line 43: @jyotigautam leave a blank line here as well.

ashish173 (approved):

PR #454, opened 2019-01-30
Test for CountryZoneMember and StateZoneMember (model/schema)

Jeshan7:
## Why?
  - Test cases were less for country and state zone member(model/schema)
  - Coverage was less for country and state zone member(model/schema)

 - Wrote new test cases for update_changset in the schema for both country zone
   and state zone member.(Schema)
 - Wrote test cases for get/1, get_all/0, member_changesets/2,
   remove_members_query/2, common_zone_query/2 for both country and
   state zone member.(Model)

[delivers #163514116]

## Checklist

### DIFF
--- a/apps/snitch_core/test/data/model/zone/country_zone_test.exs
+++ b/apps/snitch_core/test/data/model/zone/country_zone_test.exs
@@ -3,9 +3,8 @@ defmodule Snitch.Data.Model.CountryZoneTest do
   use Snitch.DataCase
 
   import Snitch.Factory
-  import Ecto.Query
-  alias Ecto.Query
 
+  alias Ecto.Query
   alias Snitch.Data.Model.CountryZone
   alias Snitch.Data.Schema.{CountryZoneMember, Zone}
 
--- a/apps/snitch_core/test/data/model/zone/country_zone_test.exs
+++ b/apps/snitch_core/test/data/model/zone/country_zone_test.exs
@@ -88,8 +88,7 @@ defmodule Snitch.Data.Model.CountryZoneTest do
     end
 
     test "fails for invalid id", %{zone: zone} do
-      {:ok, _} = CountryZone.delete(zone)
-      assert CountryZone.get(zone.id) == {:error, :zone_not_found}
+      assert {:error, :zone_not_found} = CountryZone.get(-1)
     end
   end
 
--- a/apps/snitch_core/test/data/model/zone/country_zone_test.exs
+++ b/apps/snitch_core/test/data/model/zone/country_zone_test.exs
@@ -82,66 +82,82 @@ defmodule Snitch.Data.Model.CountryZoneTest do
   end
 
   describe "get/1" do
-    test "zone with id", %{zone: zone} do
+    test "returns a zone using a zone_id", %{zone: zone} do
       {:ok, new_zone} = CountryZone.get(zone.id)
       assert new_zone.id == zone.id
+    end
+
+    test "fails for invalid id", %{zone: zone} do
+      {:ok, _} = CountryZone.delete(zone)
+      assert CountryZone.get(zone.id) == {:error, :zone_not_found}
     end
   end
 
   describe "get_all/0" do
-    @tag state_zone_count: 3,
-         country_zone_count: 3
-    test "zones  successfully", %{zones: zone} do
+    @tag country_zone_count: 3
+    test "returns all zones", %{zones: zone} do
       country_zones = CountryZone.get_all()
-      cz = Enum.drop_while(country_zones, fn x -> x.zone_type == "C" end)
-      assert cz == []
+      country_zones = Enum.all?(country_zones, fn x -> x.zone_type == "C" end)
+      assert country_zones
+    end
+
+    @tag state_zone_count: 3
+    test "fails when no country_zone is present", %{zones: zone} do
+      country_zones = CountryZone.get_all()
+      cz_list = List.delete_at(country_zones, 0)
+      country_zones = Enum.any?(cz_list, fn x -> x.zone_type == "C" end)
+      refute country_zones
     end
   end
 
   describe "member_changesets/2" do
-    test "successfully returns a CountryZoneMember changeset", %{zone: zone, countries: countries} do
+    test "returns a valid changeset", %{zone: zone, countries: countries} do
       country_ids = Enum.map(countries, &Map.get(&1, :id))
       stream = CountryZone.member_changesets(country_ids, zone)
       changeset = Enum.find_value(stream, fn x -> x end)
       assert changeset.valid?
     end
-  end
 
-  describe "remove_members_query/2" do
-    @tag country_zone_count: 3
-    test "returns a query successfully", %{zone: zone, countries: countries} do
+    test "fails to return a valid changeset", %{zone: zone, countries: countries} do
       country_ids = Enum.map(countries, &Map.get(&1, :id))
-
-      expected =
-        Query.from(c in CountryZoneMember,
-          where: c.country_id in ^country_ids and c.zone_id == ^zone.id
-        )
-
-      result = CountryZone.remove_members_query(country_ids, zone)
-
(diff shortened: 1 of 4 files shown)


apps/snitch_core/test/data/model/zone/country_zone_test.exs
```diff
@@ -77,6 +81,69 @@ defmodule Snitch.Data.Model.CountryZoneTest do
     end
   end
 
+  describe "get/1" do
+    test "zone with id", %{zone: zone} do
```
arjun289: Add the failure one too.

apps/snitch_core/test/data/model/zone/country_zone_test.exs
```diff
@@ -77,6 +81,69 @@ defmodule Snitch.Data.Model.CountryZoneTest do
...
+      {:ok, new_zone} = CountryZone.get(zone.id)
+      assert new_zone.id == zone.id
+    end
+  end
+
+  describe "get_all/0" do
+    @tag state_zone_count: 3,
```
arjun289: what's the point of setting `state_zone_count`

apps/snitch_core/test/data/model/zone/country_zone_test.exs
```diff
@@ -77,6 +81,69 @@ defmodule Snitch.Data.Model.CountryZoneTest do
...
+         country_zone_count: 3
+    test "zones  successfully", %{zones: zone} do
+      country_zones = CountryZone.get_all()
+      cz = Enum.drop_while(country_zones, fn x -> x.zone_type == "C" end)
+      assert cz == []
```
arjun289: you could have easily checked the number of countries set. Please also add test when none are set

apps/snitch_core/test/data/model/zone/country_zone_test.exs
```diff
@@ -77,6 +81,69 @@ defmodule Snitch.Data.Model.CountryZoneTest do
...
+    end
+  end
+
+  describe "member_changesets/2" do
+    test "successfully returns a CountryZoneMember changeset", %{zone: zone, countries: countries} do
```
arjun289: I think the statement should be `returns a valid changeset`

arjun289: Also, add scenario when it would be invalid.

apps/snitch_core/test/data/model/zone/country_zone_test.exs
```diff
@@ -77,6 +81,69 @@ defmodule Snitch.Data.Model.CountryZoneTest do
...
+      country_ids = Enum.map(countries, &Map.get(&1, :id))
+      stream = CountryZone.member_changesets(country_ids, zone)
+      changeset = Enum.find_value(stream, fn x -> x end)
+      assert changeset.valid?
+    end
+  end
+
+  describe "remove_members_query/2" do
+    @tag country_zone_count: 3
+    test "returns a query successfully", %{zone: zone, countries: countries} do
```
arjun289: statement should be `returns a valid query`

apps/snitch_core/test/data/model/zone/state_zone_test.exs
```diff
@@ -84,6 +88,69 @@ defmodule Snitch.Data.Model.StateZoneTest do
...
+      assert new_zone.id == zone.id
+    end
+  end
+
+  describe "get_all/0" do
+    @tag state_zone_count: 3,
+         country_zone_count: 3
+    test "zones  successfully", %{zones: zone} do
+      state_zones = StateZone.get_all()
+      sz = Enum.drop_while(state_zones, fn x -> x.zone_type == "S" end)
```
arjun289: Any particular reason you are handling it this way, I am not able to see the purpose.

apps/snitch_core/test/data/schema/zone/country_zone_member_test.exs
```diff
@@ -45,4 +60,39 @@ defmodule Snitch.Data.Schema.CountryZoneMemberTest do
...
+
+      [
+        cs:
+          %CountryZoneMember{}
+          |> CountryZoneMember.create_changeset(params)
+          |> apply_changes()
+      ]
+    end
+
+    test "successfully for valid params", %{cs: cs} do
```
arjun289: should be `returns a valid changeset` instead the statement

arjun289 (commented):
I would suggest naming the tests properly. The "describe" along with the test should return a proper statement. Also, consider not using "describe" for functions which have only one test case most likely the `get_all` ones

apps/snitch_core/test/data/model/zone/country_zone_test.exs
Jeshan7: To ensure that in the list of structs of country_zone_member and state_zone_member, the get_all/0 function retrieves only those structs which have zone_type set as "C".Similarly in the state_zone_member model also.

apps/snitch_core/test/data/model/zone/state_zone_test.exs
Jeshan7: I have used it to access all the struct records inside the list(state_zones). I have used it because only the zone_type field was common to all the records present inside the list(state_zones).

apps/snitch_core/test/data/model/zone/country_zone_test.exs
Jeshan7: In this test, I have used zones function present in the factory instead of country_zone function below. get_all/0 retrieves all the zones so while checking I have to assert the count with one extra count added to the country_zone_count tag.  
assert sz(count obtained) == 3 where country_zone_count is set to 2.
I don't think so then it would be an appropriate test because we are changing variables according to our need in the test.

arjun289: `CountryZone.get_all()` returns a list of country zones so you just need to check all the members of this list have zone type as "C". You could have used `Enum.all?(all, fn one -> one.zone_type == "C" end)`

Jeshan7: ok

apps/snitch_core/test/data/model/zone/country_zone_test.exs
```diff
@@ -77,6 +81,85 @@ defmodule Snitch.Data.Model.CountryZoneTest do
...
 
+  describe "get/1" do
+    test "returns a zone using a zone_id", %{zone: zone} do
+      {:ok, new_zone} = CountryZone.get(zone.id)
+      assert new_zone.id == zone.id
+    end
+
+    test "fails for invalid id", %{zone: zone} do
+      {:ok, _} = CountryZone.delete(zone)
+      assert CountryZone.get(zone.id) == {:error, :zone_not_found}
```
arjun289: Instead of deleting the zone you can just pass `-1` as id and it will fail

apps/snitch_core/test/data/model/zone/state_zone_test.exs
```diff
@@ -84,6 +88,85 @@ defmodule Snitch.Data.Model.StateZoneTest do
...
   end
 
+  describe "get/1" do
+    test "returns a state_zone using a zone_id", %{zone: zone} do
+      {:ok, new_zone} = StateZone.get(zone.id)
+      assert new_zone.id == zone.id
+    end
+
+    test "fails for invalid id", %{zone: zone} do
+      {:ok, _} = StateZone.delete(zone)
```
arjun289: see the comment on country `get`

arjun289 (approved):

apps/snitch_core/test/data/model/zone/country_zone_test.exs
```diff
@@ -3,11 +3,15 @@ defmodule Snitch.Data.Model.CountryZoneTest do
   use Snitch.DataCase
 
   import Snitch.Factory
+  import Ecto.Query
+  alias Ecto.Query
```
arjun289: why both `import` and `alias`? Use any one of them.

PR #468, opened 2019-02-06
Async email

jyotigautam:
## Why?
To send the email asynchronously to prevent breaking the order checkout flow.

Adding a worker under the supervisor to send order confirmation mails in the background.

[delivers #163724078 ](https://www.pivotaltracker.com/story/show/163724078)

## Checklist

### DIFF
--- a/apps/snitch_core/priv/repo/migrations/20190206091901_update_general_config_seo_title.exs
+++ b/apps/snitch_core/priv/repo/migrations/20190206091901_update_general_config_seo_title.exs
@@ -3,12 +3,9 @@ defmodule Snitch.Repo.Migrations.UpdateGeneralConfigSeoTitle do
 
   def change do
     alter table("snitch_general_configurations") do
-      remove :seo_title
-      remove :frontend_url
-      remove :backend_url
-      add :seo_title, :string, default: ""
-      add :frontend_url, :string, default: ""
-      add :backend_url, :string, default: ""
+      modify :seo_title, :string, null: true
+      modify :frontend_url, :string, null: true
+      modify :backend_url, :string, null: true
     end
   end
 end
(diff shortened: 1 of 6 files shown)


apps/snitch_core/priv/repo/migrations/20190206091901_update_general_config_seo_title.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Snitch.Repo.Migrations.UpdateGeneralConfigSeoTitle do
+  use Ecto.Migration
+
+  def change do
+    alter table("snitch_general_configurations") do
```
arjun289, line 5: What's the change here? Looks like the same thing is being added and removed.

jyotigautam: Earlier seo_title, frontend_url and backend_url didn't allow null values, but since these fields are optional I have modified the migration this way.

pkrawat1: adding default empty string

pkrawat1: so alter column should have done that, instead of removing the column. As removing the column would delete all the values in DB. So let say an `seo_title : 'Aviastore'`, will change to `seo_title: ''` What say @arjun289

jyotigautam: Yes, using [modify/3](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#modify/3) to change the field type should be the correct approach. I missed it.

pkrawat1 (approved):

